PR #2, opened 2017-11-28
Change Version string prefix to SapMachine

axel7born:
Change the product name to 'SapMachine':
```
$ java -version
sapmachine version "10"
SapMachine Runtime Environment (build 10+33)
SapMachine 64-Bit Server VM (build 10+33, mixed mode)
```


axel7born:
We don't want to replace openjdk in the version string, as some java programs explicitly check this string for something known and exit otherwise. We will add 'sapmachine' to the build number.
```
java -version
openjdk version "10"
OpenJDK Runtime Environment (build 10+33-sapmachine)
OpenJDK 64-Bit Server VM (build 10+33-sapmachine, mixed mode)
```

PR #70, opened 2018-03-15
8199557: Assert in fromTonga/vm/runtime/defmeth/scenarios/Stress_nore…

GoeLin:
Integrate from jdk11:  8199557: Assert in fromTonga/vm/runtime/defmeth/scenarios/Stress_noredefine/TestDescription.java


GoeLin:
Thanks!

PR #144, opened 2018-08-02
Add trusted CA certificate sapglobalrootca (http://www.pki.co.sap.com/)

RealCLanger:
This change adds the SAP Global Root CA certificate to the JDK trust store.


reshnm:
retest this please

PR #205, opened 2018-11-29
Processgroup support

### LINKED ISSUE #203: A minimal solution to start child processes in an own process group (*nix only)
A new switch is added, -Dsap.jdk.lang.Process.createNewProcessGroupsOnSpawn. If this switch is set to true, new processes started with ProcessBuilder.start or with Runtime.exec() will be leader of an own process group instead of belonging to the process group of the parent VM.

tstuefe:
A new switch is added, -Dsap.jdk.lang.Process.createNewProcessGroupsOnSpawn. If this switch is set to true, new processes started with ProcessBuilder.start or with Runtime.exec() will be leader of an own process group instead of belonging to the process group of the parent VM.

fixes #203


SapMachine:

tstuefe:
Hi @RealCLanger ,

> The code to parse the boolean flag in ProcessImpl.java is quite large. :) One could simply use Boolean.getBoolean()

Does that not just leak NFE if the string cannot be parsed?

> As for the naming of the flag: I suggest "sap.jdk.lang.process.createNewProcessGroupsOnSpawn". I wrote the word process with a small letter which is more what you would find in existing properties.

Ok.

> Assuming that there is a global setting for the flag given by its initial value, specified like "java -Dsap.jdk.lang.process.createNewProcessGroupsOnSpawn=true" which ought not to be changed at runtime, we can initialize this setting globally in the static initializer of ProcessGroup rather than pass it around in each call to forkAndExec(). I think it's not the design to react on global system property changes during runtime via System.setProperty()?

I do not like this, sorry. Code wise this is just as expansive as my proposal, but my proposal follows the established ways of communicating spawn attributes down to the fork point. More importantly, I want to keep the possibility of setting this property at runtime from java.

..Thomas

RealCLanger:
Hi @tstuefe 

> > The code to parse the boolean flag in ProcessImpl.java is quite large. :) One could simply use Boolean.getBoolean()
> 
> Does that not just leak NFE if the string cannot be parsed?

NFE? NumberFormatException? It's not a number but IllegalArgumentException | NullPointerException are caught and in that case false is returned. Or didn't I get your point yet? :)

> > Assuming that there is a global setting for the flag given by its initial value, specified like "java -Dsap.jdk.lang.process.createNewProcessGroupsOnSpawn=true" which ought not to be changed at runtime, we can initialize this setting globally in the static initializer of ProcessGroup rather than pass it around in each call to forkAndExec(). I think it's not the design to react on global system property changes during runtime via System.setProperty()?
> 
> I do not like this, sorry. Code wise this is just as expansive as my proposal, but my proposal follows the established ways of communicating spawn attributes down to the fork point. More importantly, I want to keep the possibility of setting this property at runtime from java.

Ok, if it's the design that the property shall be checked upon every fork then your implementation is the right thing to do 👍 

...Christoph

tstuefe:
> > > The code to parse the boolean flag in ProcessImpl.java is quite large. :) One could simply use Boolean.getBoolean()
> > 
> > 
> > Does that not just leak NFE if the string cannot be parsed?
> 
> NFE? NumberFormatException? It's not a number but IllegalArgumentException | NullPointerException are caught and in that case false is returned. Or didn't I get your point yet? :)

Yes, I meant IllegalArgumentException. But never mind, I was off. I see that Boolean.getBoolean already catches NPE and IllegalArgumentException. So, that part is fine for me.

> 
> > > Assuming that there is a global setting for the flag given by its initial value, specified like "java -Dsap.jdk.lang.process.createNewProcessGroupsOnSpawn=true" which ought not to be changed at runtime, we can initialize this setting globally in the static initializer of ProcessGroup rather than pass it around in each call to forkAndExec(). I think it's not the design to react on global system property changes during runtime via System.setProperty()?
> > 
> > 
> > I do not like this, sorry. Code wise this is just as expansive as my proposal, but my proposal follows the established ways of communicating spawn attributes down to the fork point. More importantly, I want to keep the possibility of setting this property at runtime from java.
> 
> Ok, if it's the design that the property shall be checked upon every fork then your implementation is the right thing to do 
> 

Since I cannot be sure how this feature is used, this is the only option I see.

> ...Christoph

Thanks, Thomas

tstuefe:
Withdrawn. Will do a new version.

PR #286, opened 2019-02-01
SapMachine #285: Update SapMachine specific problem lists

### LINKED ISSUE #285: Update SapMachine specific problem lists
The SapMachine specific problem lists need an update.

- Merge with the internal problem lists that SAP uses for OpenJDK regression testing
- Add SapMachine infrastructure specific exclusions to reduce the number of false positives when running platform builds

RealCLanger:
fixes #285

### DIFF
--- a/test/hotspot/jtreg/ProblemList-SapMachine.txt
+++ b/test/hotspot/jtreg/ProblemList-SapMachine.txt
@@ -26,169 +26,163 @@
 
 #############################################################################
 # Tests known to be failing in SapMachine due to SapMachine specific setup.
-#
 
-# SAPJVM GL This was in this file before 2018-10-05
+# SapMachine 2018-10-05
 # This test fails, because we do not have debug symbols available in all tests.
 runtime/NMT/CheckForProperDetailStackTrace.java                                    generic-all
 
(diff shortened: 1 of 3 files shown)


SapMachine:

RealCLanger:
Hi Goetz

> * SapMachine 2018-01-18 Failing tests after fix for 8215544. Need to check.
>   Please write instead
>   "Fixed by 8217473: SA: Tests using ClhsdbLauncher fail on SAP docker containers"

Done, thanks for the heads up. So we should be able to remove this section next week then anyway :)

> Why do you remove the solaris lines?
> -com/sun/nio/sctp/SctpChannel/Bind.java solaris-sparcv9
> I understand we don't build solaris, but it would make diffing the
> files more easy.

Ok, I went another round and included everything from the internal list also for platforms that we don't support. I agree, it makes it easier to diff the files line by line.

> We should do aother update after 7.2.2019, which is end of rdp1 for 12.
> Then we can check the OpenJDK exclude list against a stable target, maybe some
> of the issues got fixed in the meantime.

Sure, we should update this as often as possible :)

SapMachine:

SapMachine:

PR #288, opened 2019-02-01
SapMachine (11) #285: Update SapMachine specific problem lists

### LINKED ISSUE #285: Update SapMachine specific problem lists

RealCLanger:
fixes #285

### DIFF
--- a/test/hotspot/jtreg/ProblemList-SapMachine.txt
+++ b/test/hotspot/jtreg/ProblemList-SapMachine.txt
@@ -26,31 +26,87 @@
 
 #############################################################################
 # Tests known to be failing in SapMachine due to SapMachine specific setup.
-#
 
-# SapMachine 2018-10-05 This was in this file before 2018-10-05
+# SapMachine 2018-10-05
 # This test fails, because we do not have debug symbols available in all tests.
 runtime/NMT/CheckForProperDetailStackTrace.java                                    generic-all
 
-# SapMachine 2019-01-30 These tests fail for some reason on linux-ppc64le. Need to investigate.
-gc/g1/humongousObjects/TestHumongousClassLoader.java                               linux-ppc64le
-gc/g1/humongousObjects/TestHumongousNonArrayAllocation.java                        linux-ppc64le
+# SapMachine 2019-01-30 These tests fail for some reason on linux-ppc64le and linux-ppc64. Need to investigate.
+gc/g1/humongousObjects/TestHumongousClassLoader.java                               linux-ppc64le,linux-ppc64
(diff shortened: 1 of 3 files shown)


SapMachine:

SapMachine:

RealCLanger:
retest this please

SapMachine:

RealCLanger:
> I think you include the wrong variant of the OpenJDK exclude list. You need to sync back in the OpenJDK aux_repo to the tag on the corresponding file with the status of jdk11.0.2.
> In this files tests are removed that have been fixed for 11.0.3, like QuoVadis.

I took the current list which is probably at 11.03 level. There's one test excluded in SapMAchine currently which waits for a 11.03 fix. Other than that, I don't see issues in the SapMachine pr jobs and branch builds with that list. And it will evolve anyway...
 
> As with the list for 12, I would appreciate that you keep the excludes for non-SapMachine platforms to simplify comparing the files.
ok, agree. I went another round and now the common section matches line to line.

GoeLin:
For the section taken over from the openJDK jtreg exclude list it should match the right file, except for cases where you know why it needs to differ.
The jdk 11.0.2 version can be found in the internal repo with these lists:

changeset:   185:f382dfc2d63a 
user:        xxx
date:        Mon Oct 29 12:22:26 2018 +0100
summary:     Update to problem lists of tests fixed in 11.0.2

I updated the SapMachine 11 exclude list with exact this version on the day the 11.0.2 update was pulled into SapMachine 11. The only difference are the tests with the certificates that expired, there I annotated why there is a difference.  I don't see why we should now push an exclude list that does not correspond to the sources.

SapMachine:

RealCLanger:
> For the section taken over from the openJDK jtreg exclude list it should match the right file, except for cases where you know why it needs to differ.
> The jdk 11.0.2 version can be found in the internal repo with these lists:
> 
> changeset: 185:f382dfc2d63a
> user: xxx
> date: Mon Oct 29 12:22:26 2018 +0100
> summary: Update to problem lists of tests fixed in 11.0.2
> 
> I updated the SapMachine 11 exclude list with exact this version on the day the 11.0.2 update was pulled into SapMachine 11. The only difference are the tests with the certificates that expired, there I annotated why there is a difference. I don't see why we should now push an exclude list that does not correspond to the sources.

as discussed 1on1, I've added back the missing QuoVadis test. Now the list should more or less ressemble the 11.0.2 state in the internal repo. Starting with 11.03 we should ge back in sync then.

PR #305, opened 2019-02-26
Initial implementation for #241 - A low-cost, always on statistical v…

### LINKED ISSUE #241: A low-cost, always on statistical value history (JDK-8212618)
This issue tracks development of [JDK-8212618 "A low-cost, always on statistical value history"](https://bugs.openjdk.java.net/browse/JDK-8212618). 

This proposed feature has been rejected upstream since it is perceived to collide with JMC/JFR.
We consider it useful enough and the merge cost to be small enough to integrate it into the SapMachine. 

Details:

SAP developed a supportability feature called "Statistics History" which is a low cost history of statistical values-of-interest. 

These values contain parameters of the jvm (e.g. heap size, metaspace size, number of loaded classes) and the underlying platform (e.g. rss, swapping state, run queue length etc). At intervals of (by default) 60 seconds these values are measured and stored in a fifo buffer. 

The fifo buffer has three parts, a short-, medium-, and long term fifo buffer. A fraction of the samples falling out of the short term fifo is transferred to the mid term fifo; again, a fraction of the samples falling out of the mid term fifo is transferred to the long term fifo. So, the short term fifo covers a short recent timespan (usually an hour) in comparativly short sample intervals (usually 60 seconds), whereas the long term fifo covers a very long time span (~10 days) with interval times of hours. 

This feature has been very popular with our support folks and so we would like to contribute that. It enables us to easily analyze slowly developing situations like memory leaks, memory or cpu spikes, resource starvation etc. 

[\[1\] JDK-8212618 - Original JBS issue, closed as Wont Fix](https://bugs.openjdk.java.net/browse/JDK-8212618)
[\[2\] Patch on Github](https://github.com/tstuefe/ojdk-stathist-patch)
[\[3\] Example output](http://cr.openjdk.java.net/~stuefe/webrevs/stathist/examples/dacapo.txt)

tstuefe:
…alue history (JDK-8212618)

fixes #241

### DIFF
--- a/src/hotspot/os/linux/stathist_linux.cpp
+++ b/src/hotspot/os/linux/stathist_linux.cpp
@@ -1,0 +1,415 @@
+/*
+ * Copyright (c) 2019, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, SAP SE.
+ *
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ *
+ */
+
+#include "precompiled.hpp"
+#include "runtime/os.hpp"
+#include "services/stathist_internals.hpp"
+#include "utilities/debug.hpp"
+#include "utilities/globalDefinitions.hpp"
+
+#include <fcntl.h>
+#include <string.h>
+#include <errno.h>
+
+namespace StatisticsHistory {
+
+class ProcFile {
+  char* _buf;
+
+  // To keep the code simple, I just use a fixed sized buffer.
+  enum { bufsize = 4*K };
+
+public:
+
+  ProcFile() : _buf(NULL) {
+    _buf = (char*)os::malloc(bufsize, mtInternal);
+  }
+
+  ~ProcFile () {
+    os::free(_buf);
+  }
+
+  bool read(const char* filename) {
+
+    FILE* f = ::fopen(filename, "r");
+    if (f == NULL) {
+      return false;
+    }
+
+    size_t bytes_read = ::fread(_buf, 1, bufsize, f);
+    _buf[bufsize - 1] = '\0';
+
+    ::fclose(f);
+
+    return bytes_read > 0 && bytes_read < bufsize;
+  }
+
+  const char* text() const { return _buf; }
+
+  const char* get_prefixed_line(const char* prefix) const {
+    const char* p = ::strstr(_buf, prefix);
+    if (p != NULL) {
+      return p;
+    }
+    return NULL;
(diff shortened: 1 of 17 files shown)


SapMachine:
Hi @tstuefe. Thanks for your Pull Request. I'm waiting for a [SapMachine Team member](https://github.com/orgs/SAP/teams/sapmachine-team) to verify that this pull request is ok to test. If it is, they should reply with `retest this please`.

ArnoZeller:
retest this please

SapMachine:

SapMachine:

tstuefe:
> Hi Thomas,
> 
> I looked at it only from an integration point of view. From that end, I don't see an issue merging this into the sapmachine codebase. Thanks for this work :)
> 
> As for the test: SapMachine pr validation tests only test tier1. Your new test, however, isn't part of tier1 yet. Can you please add it to tier 1 by introducing a tier1_sapmachine group in the hotspot jtreg tests. You can take this commit as an example (where I did it for jdk jtreg):
> [cb905d5#diff-ac915910d0d576b0fb8f40dcb0abb3df](https://github.com/SAP/SapMachine/commit/cb905d5358ec143834fa6f171680ad18add73ec1#diff-ac915910d0d576b0fb8f40dcb0abb3df)
> 
> Best regards
> Christoph

Hi Christoph! Thanks. Done.

..Thomas

tstuefe:
ping..

src/hotspot/os/linux/stathist_linux.cpp
```diff
@@ -0,0 +1,415 @@
...
+
+    ::fclose(f);
+
+    return bytes_read > 0 && bytes_read < bufsize;
+  }
+
+  const char* text() const { return _buf; }
+
+  const char* get_prefixed_line(const char* prefix) const {
+    const char* p = ::strstr(_buf, prefix);
```
schmelter-sap: replace code by
`return ::strstr(_buf, prefix);`

src/hotspot/os/linux/stathist_linux.cpp
```diff
@@ -0,0 +1,415 @@
...
+    return NULL;
+  }
+
+  value_t parsed_prefixed_value(const char* prefix, size_t scale = 1) const {
+    value_t value = INVALID_VALUE;
+    const char* const s = get_prefixed_line(prefix);
+    if (s != NULL) {
+      errno = 0;
+      const char* p = s + ::strlen(prefix);
+      value = (value_t)::strtoll(p, NULL, 10);
```
schmelter-sap: Use the second arg to stroll to check if the value was really 0 or invalid.

src/hotspot/os/linux/stathist_linux.cpp
```diff
@@ -0,0 +1,415 @@
...
+        // Values are in ticks. Convert to ms.
+        const uint64_t value_ms = (value * 1000) / _clk_tck;
+        const uint64_t last_value_ms = (last_value * 1000) / _clk_tck;
+        const uint64_t delta_ms = value_ms - last_value_ms;
+
+        // Calculate the number of wallclock milliseconds for the delta interval...
+        const int age_ms = last_value_age * 1000;
+
+        // times number of available cores.
+        const int total_cpu_time_ms = age_ms * _num_cores;
```
schmelter-sap: age_ms and total_cpu_time_ms are only ints. If you are on a machine with many cores (say 512) the last value can easily overflow for last_value_age values > 1h.

src/hotspot/os/linux/stathist_linux.cpp
```diff
@@ -0,0 +1,415 @@
...
+
+        // Put the spent cpu time in reference to the total available cpu time.
+        const float percentage = (100.0f * delta_ms) / total_cpu_time_ms;
```
schmelter-sap: Why float and not double? It will be converted to double in the jio_sprintf call anyways.

src/hotspot/share/services/stathist.cpp
```diff
@@ -0,0 +1,1192 @@
...
+  struct tm _tm;
+  if (os::localtime_pd(&t, &_tm) == &_tm) {
+    char buf[32];
+    ::strftime(buf, sizeof(buf), "%Y-%m-%d %H:%M:%S", &_tm);
+    st->print("%*s", TIMESTAMP_LEN, buf);
+  }
+}
+
+// A little RAII class to temporarily change locale to "C"
+class CLocaleMark {
```
schmelter-sap: This is not threadsafe since it is a global configuration. A parallel running thread which runs locale sensitive code would be effected.

PR #421, opened 2019-06-10
Rename external representation of Statistical History

### LINKED ISSUE #420: Rename Statistical History to Vitals [History]
Because of aesthetics and because "Statistical History" is not exactly rolling off the tongue. Change external representation of this feature from "Statistical History" aka "StatHist" to "Vitals History" or "Vitals":

-  VM.stathist -> VM.vitals
-  -XX:+EnableStatHist -> -XX:+EnableVitals

To keep the change to a minimum and to keep review work low, rename only the externals. Leave class names etc as they were.

tstuefe:
Rename "Statistical History" to "Vitals" with a minimum of fuss:
1) rename command: VM.stathist -> VM.vitals
2) rename switches: "EnableStatHist" -> "EnableVitals" etc.

fixes #420

### DIFF
--- a/src/hotspot/share/runtime/globals.hpp
+++ b/src/hotspot/share/runtime/globals.hpp
@@ -514,7 +514,8 @@ const size_t minimumSymbolTableSize = 1024;
                                                                             \
   /* SapMachine 2019-02-20 : stathist */                                    \
   product(bool, EnableVitals, true,                                         \
-          "Enable recording of vitals")                                     \
+          "Enable sampling of vitals: memory, cpu utilization and various " \
+          "VM core statistics; display via jcmd \"VM.vitals\".")            \
                                                                             \
   product(uintx, VitalsSampleInterval, 0,                                   \
           "Vitals sample rate interval (0=use default sample rate)")        \
--- a/src/hotspot/share/runtime/globals.hpp
+++ b/src/hotspot/share/runtime/globals.hpp
@@ -513,14 +513,14 @@ const size_t minimumSymbolTableSize = 1024;
           "Print additional debugging information from other modes")        \
                                                                             \
   /* SapMachine 2019-02-20 : stathist */                                    \
-  product(bool, EnableStatHist, true,                                       \
-          "Enable Statistics history")                                      \
+  product(bool, EnableVitals, true,                                         \
+          "Enable recording of vitals")                                     \
                                                                             \
-  product(uintx, StatHistSampleInterval, 0,                                 \
-          "Statistics history sample rate interval (0=default)")            \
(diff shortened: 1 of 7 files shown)


SapMachine:

tstuefe:
> Looks good.
> 

Thanks.

> When merging, please squash and change commit message to:
> "SapMachine #420: Rename external representation of Statistical History"
> 

okay.

> I think we should also backport this to 11 and 12 for the sake of consistency, shouldn't we?

Sure, thats the plan :)

src/hotspot/share/runtime/globals.hpp
```diff
@@ -513,14 +513,14 @@ const size_t minimumSymbolTableSize = 1024;
           "Print additional debugging information from other modes")        \
                                                                             \
   /* SapMachine 2019-02-20 : stathist */                                    \
-  product(bool, EnableStatHist, true,                                       \
-          "Enable Statistics history")                                      \
+  product(bool, EnableVitals, true,                                         \
+          "Enable recording of vitals")                                     \
```
GoeLin: * I would use the same word as below: "Enable __sampling__ of vitals"
  (or is the sampling on even if this flag is off?)
* I would give a short explanation of what vitals are:
"Enable sampling of vitals: memory, heap and cpu utilization" (Please use your own text ... this is only intended as example).

src/hotspot/share/services/stathist.cpp
```diff
@@ -1132,7 +1132,7 @@ void cleanup() {
 
 void print_report(outputStream* st, const print_info_t* pi) {
 
-  st->print("stathist:");
+  st->print("Vitals History:");
```
GoeLin: I would omit the word "History".  As I understand, there is no printout of a single sample which would justify to distinguish "Vitals" and "Vitals History"?

src/hotspot/share/services/stathistDCmd.hpp
```diff
@@ -41,10 +41,10 @@ class StatHistDCmd : public DCmdWithParser {
 public:
   StatHistDCmd(outputStream* output, bool heap);
   static const char* name() {
-    return "VM.stathist";
+    return "VM.vitals";
   }
   static const char* description() {
-    return "Provide statistics history.";
+    return "Print Vitals History.";
```
GoeLin: omit "History" too.

SapMachine:

src/hotspot/share/runtime/globals.hpp
tstuefe: Thanks Goetz. All good proposals. Will post update.

SapMachine:

PR #424, opened 2019-06-15
Merge to tag jdk-13+25

SapMachine:
please pull


RealCLanger:
I opened Bug JDK-8226302 (https://bugs.openjdk.java.net/browse/JDK-8226302) and excluded the failing tests on ppc via the SapMachine specific exclusion list. Let's see if the rebuild is green and whether the macos issue can be observed again.

PR #447, opened 2019-07-11
Merge to tag jdk-13+29

SapMachine:
please pull


RealCLanger:
I merged in sapmachine13, removing the test exclusions after JDK-8226302 was resolved. Let's see if test failures reappear... (and the build problem on linuxx86_64)

RealCLanger:
As for the build error, I opened https://bugs.openjdk.java.net/browse/JDK-8227578 and proposed a patch. Looks like we're good to merge now.

PR #504, opened 2019-09-19
Sapmachine (11) #492: Various improvements to vitals

### LINKED ISSUE #492: Various improvements to vitals
This is an umbrella issue for a number of improvements to the *vitals* feature:

1) consistently name the "comma-separated-value" output feature "csv" (it was erroneously named "cvs" in the initial commit).

2) Make it possible to reverse the printing order in the historic tables. Now, the default is we print the samples youngest-to-oldest. That is inconvenient, especially for processing in csv form. Add an option to VM.vitals to optionally reverse that order.

3) When printing comma-separated-values, enclose them in quotes. That prevents localization issues if LC_NUMERIC is not en_US.

4) The "scale" option of VM.vitals should accept bytes, aka 1, to get precise values for automated processing. This would also be consistent with what NMT does.

5) A new option should be added, "-XX:[+-]DumpVitalsAtExit" (off by default) which should dump the vitals history both in textual and csv form when the VM exits.

6) The existing "raw" output option for VM.vitals is currently only enabled on debug builds. However, it is useful in general and I would like to have it in release builds as well.

(Note: all this is supposed to be backported to our older releases).

RealCLanger:
SapMachine #492: Various improvements to vitals

(cherry picked from commit edf4899dcafbee27bc1aba67d635dfda0ace15e4)

fixes #492

### DIFF
--- a/src/hotspot/share/runtime/globals.hpp
+++ b/src/hotspot/share/runtime/globals.hpp
@@ -610,7 +610,7 @@ define_pd_global(uint64_t,MaxRAM,                    1ULL*G);
   develop(bool, Verbose, false,                                             \
           "Print additional debugging information from other modes")        \
                                                                             \
-  /* SapMachine 2019-02-20 : stathist */                                    \
+  /* SapMachine 2019-02-20 : vitals */                                      \
   product(bool, EnableVitals, true,                                         \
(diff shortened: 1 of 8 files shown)


SapMachine:

PR #513, opened 2019-09-28
SapMachine #512: Make file name for DumpVitalsAtExit configurable

### LINKED ISSUE #512: Make file name for DumpVitalsAtExit configurable
For automated scripting it makes sense to be able to set the output file name for the vitals dumped at vm exit.

tstuefe:
fixes #512

### DIFF
--- a/src/hotspot/share/runtime/globals.hpp
+++ b/src/hotspot/share/runtime/globals.hpp
@@ -514,8 +514,13 @@ const size_t minimumSymbolTableSize = 1024;
           "When sampling vitals, omit any actions which require locking.")  \
                                                                             \
   product(bool, DumpVitalsAtExit, false,                                    \
-	        "Dump vitals at VM exit into two files, vitals_<pid>.txt and "    \
-	        "vitals_<pid>.csv, respectively.")                                \
+	        "Dump vitals at VM exit into two files, by default called "       \
+	        "sapmachine_vitals_<pid>.txt and sapmachine_vitals_<pid>.csv. "   \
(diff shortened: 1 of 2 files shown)


SapMachine:

RealCLanger:
Hey Thomas,
I think this is useful. It would be nice though if one could specify a placeholder that gets substituted with the pid in the new filename option, e.g. <pid>. What do you think?
Best regards 
Christoph

tstuefe:
> Hey Thomas,
> I think this is useful. It would be nice though if one could specify a placeholder that gets substituted with the pid in the new filename option, e.g. . What do you think?
> Best regards
> Christoph

Hi Christoph,

Okay, but can we just merge this one and add placeholders later if needed?

Personally, I find the "pid" placeholder is not used much in reality. The point of this particular pull request is exactly to avoid the "randomness" of a pid in the name, to simplify automatic processing.

Cheers, Thomas

SapMachine:

RealCLanger:
Hi Thomas,

> Okay, but can we just merge this one and add placeholders later if needed?
> 
> Personally, I find the "pid" placeholder is not used much in reality. The point of this particular pull request is exactly to avoid the "randomness" of a pid in the name, to simplify automatic processing.

It was just an idea. But maybe you're right. For scripting you don't want the pid and for other use cases such as a server that is restarted recently one would probably not use the pid placeholder as well since otherwise over time a bunch of such files would accumulate. So let's forget about this until somebody asks for it with a strong usecase :)

/Christoph

PR #541, opened 2019-11-07
SapMachine #540: Various fixes and improvements to Vitals

### LINKED ISSUE #540: Various fixes and improvements to Vitals
- Fix build problem on gcc 5.4 due to using enum as constants.
  "stathist.cpp:885:63: error: enumeral and non-enumeral type in conditional expression [-Werror=extra]"
- Add a safety net for suspected endless looping in printing code. I could not reproduce it, nor could I spot any error in my iterator code, but we have some rare cases which look like the iterator code may be hanging.
- Removed coding to take a sample at printing time. This was not used and is potentially dangerous (some sampled values need locks, and we do not want to lock during printing).
- Reduce sample interval to 10 seconds (from 15)
- Added a new switch to VM.metaspace, "max", which limits the number of records printed. Together with "no-legend", "max=1" is an easy way to obtain life vital values from a number of running machines.
- Add Vitals sampler thread to thread lists in hs-err file and jcmd Thread.print

tstuefe:
Various fixes and improvements to Vitals
fixes #540

### DIFF
--- a/src/hotspot/share/services/diagnosticCommand.cpp
+++ b/src/hotspot/share/services/diagnosticCommand.cpp
@@ -144,7 +144,6 @@ void DCmdRegistrant::register_dcmds_ext(){
 }
 #endif
 
-
 HelpDCmd::HelpDCmd(outputStream* output, bool heap) : DCmdWithParser(output, heap),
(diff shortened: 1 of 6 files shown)


SapMachine:

SapMachine:

SapMachine:

PR #545, opened 2019-11-14
SapMachine (11) #540: Various fixes and improvements to vitals

### LINKED ISSUE #540: Various fixes and improvements to Vitals

RealCLanger:
SapMachine #540: Various fixes and improvements to vitals

* Fix build error on gcc 5.4, Ubuntu 16.4
"stathist.cpp:885:63: error: enumeral and non-enumeral type in conditional expression [-Werror=extra]"

* Add safety measure to guard against suspected endless loop in printing code.

* Remove the take-sample-at-printing-time feature since it
could be dangerous and the output was confusing.

* Reduce default sample interval from 15 to 10 seconds.

* Add a new sub option, "max", to limit number of records printed.

* Add Vitals sampler thread to threads list output.

(cherry picked from commit ff04966519a91608129811c2aa94a0c14fe6a328)

fixes #540

### DIFF
--- a/src/hotspot/share/runtime/thread.cpp
+++ b/src/hotspot/share/runtime/thread.cpp
@@ -4630,6 +4630,13 @@ void Threads::print_on(outputStream* st, bool print_stacks,
     st->cr();
   }
 
+  // SapMachine 2019-11-07 : stathist
+  const Thread* stathist_sampler_thread = StatisticsHistory::samplerthread();
+  if (stathist_sampler_thread != NULL) {
+    stathist_sampler_thread->print_on(st);
+    st->cr();
+  }
+
   st->flush();
 }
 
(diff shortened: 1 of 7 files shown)


SapMachine:

RealCLanger:
I think this can be merged. Windows has infrastructure issues currently.

PR #551, opened 2019-11-21
Merge to tag jdk-11.0.6+4

SapMachine:
please pull


RealCLanger:
This one can be merged, too. The MacOS error needs to be addressed separately and Windows failed due to infra shakyness.

ArnoZeller:
Mac is a know problem on Catalina - Windows looks to me like a shaky test that did hang and then time-outed causing problems during cleanup.
Will retry windows.

PR #623, opened 2020-03-19
Merge to tag jdk-15+15

SapMachine:
please pull


RealCLanger:
ppc64 issue is an intermittent thing.
MacOS problems could be related to https://bugs.openjdk.java.net/browse/JDK-8238268. We'll merge nevertheless and investigate later.

PR #638, opened 2020-04-16
Merge to tag jdk-15+19

SapMachine:
please pull


RealCLanger:
3 PR validation jobs failed:
ppc64: java/math/BigInteger/largeMemory/SymmetricRangeTests.java -> intermittent
ppc64le: runtime/CompressedOops/CompressedClassPointers.java -> we see this lately, need to analyze
mac: hotspot SA tests fail, we probably need to check/reboot mac node.

After all, we should merge this.

PR #655, opened 2020-06-11
Merge to tag jdk-11.0.8+7

SapMachine:
please pull


RealCLanger:
Seems Windows test failure is known issue, as per https://bugs.openjdk.java.net/browse/JDK-8246387. We have to think of a solution for 11u

PR #683, opened 2020-07-08
Merge to tag jdk-16+4

SapMachine:
please pull


RealCLanger:
I think that's good to merge. MacOS didn't build because of infrastructure issues.

PR #816, opened 2021-02-05
Merge to tag jdk-17+8

SapMachine:
please pull


RealCLanger:
I think this is good.

The failing test on Mac is due to a new build agent with Big Sur. The issue is already reported in OpenJDK with https://bugs.openjdk.java.net/browse/JDK-8253702 (for which I'm trying to find a fix now). Anyway, the issue that os.version=10.16 on Big Sur was always present and just didn't manifest in the test run.

linux ppc64 is an intermittent item, no reason to hold back this PR for it.

ArnoZeller (approved):
macOS test issue is not relevant.
ppc64 issue is caused by native OOM - no test issue.

Looks good to me.

RealCLanger (approved):

PR #819, opened 2021-02-11
Merge to tag jdk-17+9

SapMachine:
please pull


ArnoZeller (approved):
Looks acceptable to me.
Small CDS issue on LinuxPPC and timeouts on Mac that might be investigated, But seems to be tolerable -:)

RealCLanger (approved):
Let's merge this, given that it is a JDK17 weekstone and test issues only show up in ppc64 and new MacOS system. Should be analyzed/fixed, though.

PR #823, opened 2021-02-18
Merge to tag jdk-17+10

SapMachine:
please pull


RealCLanger:
Mac and linuxppc64 have a few test failures where we should have a look at...

ArnoZeller:
linux_ppc64 seems ok - some tests seem to expect cores, but the coresize is 0. Might by a configuration issue. Need to take a look later. No showstopper.
macos_x86_64 is also ok: the two test needs investigation - but as both timeouted it looks like an issue with load on the test machine ...

ArnoZeller (approved):
Looks ok to me.
The test issues seem to be no showstoppers but machine specific,

RealCLanger (approved):

PR #827, opened 2021-03-03
Merge to tag jdk-17+12

SapMachine:
please pull


ArnoZeller (approved):
Looks acceptable to me. 
We could merge but should investigate the PPC issues further. 
It seems they are related to test setup issues (we do not see them on other machines). But it might be that there are real issues behind.

RealCLanger (approved):
Agree, let's merge this but check the ppc issues.

PR #829, opened 2021-03-11
Merge to tag jdk-17+13

SapMachine:
please pull


ArnoZeller (approved):
hotspot - serviceability/dcmd/gc/RunFinalizationTest.java: Test did run in timeout on aarch64 - seems to happen sporadic on different plattforms (https://bugs.openjdk.java.net/browse/JDK-8227120)
jdk - java/math/BigInteger/largeMemory/SymmetricRangeTests.java: Test was killed on both LinuxPPC plattformsby OOM killer - will try to enable swapfile to avoid this issue.
htospot - gc/shenandoah/oom/TestClassLoaderLeak.java: Failed on macOS- no showstopper but we should keep an eye on it.

RealCLanger (approved):
Thanks for having a look at the test issues, Arno. I'll go ahead and merge then. Let's see if we see the Mac test failure again.

PR #853, opened 2021-04-29
Merge to tag jdk-17+20

SapMachine:
please pull


ArnoZeller (approved):
Looks good - only issues found in testing is langtools/jdk/jshell/ToolBasicTest.java on macOS that has locale problems.
We better check the locale on the test machine or improve the test.

RealCLanger (approved):

PR #855, opened 2021-05-06
Merge to tag jdk-11.0.12+1

SapMachine:
please pull


ArnoZeller (approved):
Several jshell tests fail on linux aarch64 - but as we still ship aarch only as beta this is no showstopper.

PR #862, opened 2021-05-21
Unify behavior for xxxOnOutOfMemoryError switches

### LINKED ISSUE #861: Unify behavior for xxxOnOutOfMemoryError switches
Behavior when handling xxxOnOutOfMemoryError switches should be unified between SAPJVM and SapMachine.

The following changes should be made:
1) A switch `ExitVMOnOutOfMemoryError` should be introduced as an alias to `CrashOnOutOfMemoryError`. This switch exists for historical reasons in the SAPJVM and should exist in the SapMachine too.
2) The behavior of  `CrashOnOutOfMemoryError` should change as follows:
   - no core dump should be written unless CreateCoreDumpOnCrash has been *explicitly* enabled on the command line
   - a thread stack of the OOMing Java thread should be written to stdout
3) The behavior of `ExitOnOutOfMemoryError` (note, not Exit**VM**OnOutOfMemoryError) changes:
   - a thread stack of the OOMing Java thread should be written to stdout
4) All xxxOnOutOfMemoryError switches (`CrashOnOutOfMemoryError`, `ExitVMOnOutOfMemoryError`, `ExitOnOutOfMemoryError` and `HeapDumpOnOutOfMemoryError`) should also work for OOMs caused by the inability of the platform to start threads.

(Note: I tried to bring (3) upstream but failed. We may try this again later, but for now, we are willing to accept that SapMachine will show a slightly different behavior than stock OpenJDK)

tstuefe:
Hi, may I please have reviews for this change. Details and motivation see https://github.com/SAP/SapMachine/issues/861.

The patch:

- introduces `ExitVMOnOutOfMemoryError` as an alias to `CrashOnOutOfMemoryError`
- adds stack dumping to tty for both
- switches off cores if either one is set, unless cores had been explicitly requested by the user
- Wires all `xxOnOutOfMemoryError` switches up to thread creation too - all these will fire when we cannot start native threads (note that `OnOutOfMemoryError=command` probably won't work since it requires the ability to fork). Note that I decided to wire this up after resource-exhausted-JVMTI-hook is called; does that make sense or should I do it before? to me its a coin toss, really.)
- Adds a new jtreg test to test SapMachine specific logic
- Also adds stack dumping to the otherwise unchanged `ExitOnOutOfMemoryError`
---

Testing:
- ran the jtreg test
- ran hotspot:tier1 locally
- tested manually all these switches both with OOM from Metaspace and OOMs from thread creation; the latter with ulimit -u set. Works. We now print thread stack to tty, then assert, but without core:

```
thomas@starfish$ ulimit -u 3000
thomas@starfish$ ./images/jdk/bin/java -XX:+CrashOnOutOfMemoryError de.stuefe.repros.MultiThreadTest --num-threads=3000 --wait-time=100
Will start 3000 threads, wait time 100s...
<press key>
 ... continuing.
Created: 300...
Created: 600...
Created: 900...
Created: 1200...
Created: 1500...
Created: 1800...
Created: 2100...
[1,119s][warning][os,thread] Failed to start thread - pthread_create failed (EAGAIN) for attributes: stacksize: 1024k, guardsize: 0k, detached.
Aborting due to java.lang.OutOfMemoryError: unable to create native thread: possibly out of memory or process/resource limits reached
Java frames: (J=compiled Java code, j=interpreted, Vv=VM code)
J 372  java.lang.Thread.start0()V java.base@17-internal (0 bytes) @ 0x00007fe389087ac4 [0x00007fe389087a60+0x0000000000000064]
J 368 c1 java.lang.Thread.start()V java.base@17-internal (71 bytes) @ 0x00007fe381cd0934 [0x00007fe381cd07e0+0x0000000000000154]
j  de.stuefe.repros.MultiThreadTest.call()Ljava/lang/Integer;+93
j  de.stuefe.repros.MultiThreadTest.call()Ljava/lang/Object;+1
j  picocli.CommandLine.executeUserObject(Lpicocli/CommandLine;Ljava/util/List;)Ljava/util/List;+94
j  picocli.CommandLine.access$900(Lpicocli/CommandLine;Ljava/util/List;)Ljava/util/List;+2
j  picocli.CommandLine$RunLast.handle(Lpicocli/CommandLine$ParseResult;)Ljava/util/List;+29
j  picocli.CommandLine$RunLast.handle(Lpicocli/CommandLine$ParseResult;)Ljava/lang/Object;+2
j  picocli.CommandLine$AbstractParseResultHandler.execute(Lpicocli/CommandLine$ParseResult;)I+16
j  picocli.CommandLine.execute([Ljava/lang/String;)I+31
j  de.stuefe.repros.MultiThreadTest.main([Ljava/lang/String;)V+15
v  ~StubRoutines::call_stub
# To suppress the following error report, specify this argument
# after -XX: or in .hotspotrc:  SuppressErrorAt=/debug.cpp:366
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  Internal Error (/shared/projects/openjdk/sapmachine/source/src/hotspot/share/utilities/debug.cpp:366), pid=441918, tid=441919
#  fatal error: OutOfMemory encountered: unable to create native thread: possibly out of memory or process/resource limits reached
#
# JRE version: OpenJDK Runtime Environment (17.0) (fastdebug build 17-internal+0-adhoc.thomas.source)
# Java VM: OpenJDK 64-Bit Server VM (fastdebug 17-internal+0-adhoc.thomas.source, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, linux-amd64)
# CreateCoredumpOnCrash turned off, no core file dumped
#
# An error report file with more information is saved as:
# /shared/projects/openjdk/sapmachine/output-fastdebug/hs_err_pid441918.log

```

fixes #861

### DIFF
--- a/src/hotspot/share/prims/jvm.cpp
+++ b/src/hotspot/share/prims/jvm.cpp
@@ -2908,15 +2908,16 @@ JVM_ENTRY(void, JVM_StartThread(JNIEnv* env, jobject jthread))
   if (native_thread->osthread() == NULL) {
     // No one should hold a reference to the 'native_thread'.
     native_thread->smr_delete();
+
+    // SapMachine 2021-05-21: All ...OnOutOfMemoryError switches should work for
+    //  thread creation failures too.
+    report_java_out_of_memory(os::native_thread_creation_failed_msg());
+
     if (JvmtiExport::should_post_resource_exhausted()) {
       JvmtiExport::post_resource_exhausted(
         JVMTI_RESOURCE_EXHAUSTED_OOM_ERROR | JVMTI_RESOURCE_EXHAUSTED_THREADS,
         os::native_thread_creation_failed_msg());
     }
-
-    // SapMachine 2021-05-21: All ...OnOutOfMemoryError switches should work for
-    //  thread creation failures too.
-    report_java_out_of_memory(os::native_thread_creation_failed_msg());
 
     THROW_MSG(vmSymbols::java_lang_OutOfMemoryError(),
               os::native_thread_creation_failed_msg());
--- a/src/hotspot/share/prims/jvm.cpp
+++ b/src/hotspot/share/prims/jvm.cpp
@@ -2918,7 +2918,6 @@ JVM_ENTRY(void, JVM_StartThread(JNIEnv* env, jobject jthread))
         JVMTI_RESOURCE_EXHAUSTED_OOM_ERROR | JVMTI_RESOURCE_EXHAUSTED_THREADS,
         os::native_thread_creation_failed_msg());
     }
-
     THROW_MSG(vmSymbols::java_lang_OutOfMemoryError(),
               os::native_thread_creation_failed_msg());
   }
--- a/src/hotspot/share/prims/jvm.cpp
+++ b/src/hotspot/share/prims/jvm.cpp
@@ -2913,6 +2913,11 @@ JVM_ENTRY(void, JVM_StartThread(JNIEnv* env, jobject jthread))
         JVMTI_RESOURCE_EXHAUSTED_OOM_ERROR | JVMTI_RESOURCE_EXHAUSTED_THREADS,
         os::native_thread_creation_failed_msg());
     }
+
+    // SapMachine 2021-05-21: All ...OnOutOfMemoryError switches should work for
+    //  thread creation failures too.
+    report_java_out_of_memory(os::native_thread_creation_failed_msg());
+
     THROW_MSG(vmSymbols::java_lang_OutOfMemoryError(),
               os::native_thread_creation_failed_msg());
   }
--- a/src/hotspot/share/runtime/arguments.cpp
+++ b/src/hotspot/share/runtime/arguments.cpp
@@ -3146,8 +3146,8 @@ jint Arguments::finalize_vm_init_args(bool patch_mod_javabase) {
 
   // SapMachine 2021-05-21:
   // - ExitVMOnOutOfMemory is an alias for CrashOnOutOfMemoryError
-  // - if either one of those is specified, and Crash dumps have not been explicitly enabled,
-  //   turn dumps off (we don't want cores on OOM)
+  // - if either one of these two is active and Crash dumps had not been explicitly enabled,
+  //   we turn dumps off (we generally don't want cores on OOM)
   if (ExitVMOnOutOfMemoryError && !CrashOnOutOfMemoryError) {
     FLAG_SET_ERGO(CrashOnOutOfMemoryError, true);
   }
--- a/src/hotspot/share/runtime/arguments.cpp
+++ b/src/hotspot/share/runtime/arguments.cpp
@@ -3146,7 +3146,7 @@ jint Arguments::finalize_vm_init_args(bool patch_mod_javabase) {
 
   // SapMachine 2021-05-21:
   // - ExitVMOnOutOfMemory is an alias for CrashOnOutOfMemoryError
-  // - if either one of those is specified, and Crash dumps have not been explicitly disabled,
+  // - if either one of those is specified, and Crash dumps have not been explicitly enabled,
   //   turn dumps off (we don't want cores on OOM)
   if (ExitVMOnOutOfMemoryError && !CrashOnOutOfMemoryError) {
     FLAG_SET_ERGO(CrashOnOutOfMemoryError, true);
--- a/src/hotspot/share/runtime/arguments.cpp
+++ b/src/hotspot/share/runtime/arguments.cpp
@@ -3144,15 +3144,9 @@ jint Arguments::finalize_vm_init_args(bool patch_mod_javabase) {
   UNSUPPORTED_OPTION(ShowRegistersOnAssert);
 #endif // CAN_SHOW_REGISTERS_ON_ASSERT
 
-  // SapMachine 2021-05-21:
-  // - ExitVMOnOutOfMemory is an alias for CrashOnOutOfMemoryError
-  // - if either one of these two is active and Crash dumps had not been explicitly enabled,
-  //   we turn dumps off (we generally don't want cores on OOM)
+  // SapMachine 2021-05-21: Let ExitVMOnOutOfMemory be an alias for CrashOnOutOfMemoryError
   if (ExitVMOnOutOfMemoryError && !CrashOnOutOfMemoryError) {
     FLAG_SET_ERGO(CrashOnOutOfMemoryError, true);
-  }
-  if (CrashOnOutOfMemoryError && CreateCoredumpOnCrash && FLAG_IS_DEFAULT(CreateCoredumpOnCrash)) {
-    FLAG_SET_ERGO(CreateCoredumpOnCrash, false);
   }
 
   return JNI_OK;
--- a/src/hotspot/share/runtime/arguments.cpp
+++ b/src/hotspot/share/runtime/arguments.cpp
@@ -3144,6 +3144,17 @@ jint Arguments::finalize_vm_init_args(bool patch_mod_javabase) {
   UNSUPPORTED_OPTION(ShowRegistersOnAssert);
 #endif // CAN_SHOW_REGISTERS_ON_ASSERT
 
+  // SapMachine 2021-05-21:
+  // - ExitVMOnOutOfMemory is an alias for CrashOnOutOfMemoryError
+  // - if either one of those is specified, and Crash dumps have not been explicitly disabled,
+  //   turn dumps off (we don't want cores on OOM)
+  if (ExitVMOnOutOfMemoryError && !CrashOnOutOfMemoryError) {
+    FLAG_SET_ERGO(CrashOnOutOfMemoryError, true);
+  }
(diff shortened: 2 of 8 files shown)


SapMachine:

src/hotspot/share/runtime/arguments.cpp
```diff
@@ -3144,6 +3144,17 @@ jint Arguments::finalize_vm_init_args(bool patch_mod_javabase) {
   UNSUPPORTED_OPTION(ShowRegistersOnAssert);
 #endif // CAN_SHOW_REGISTERS_ON_ASSERT
 
+  // SapMachine 2021-05-21:
+  // - ExitVMOnOutOfMemory is an alias for CrashOnOutOfMemoryError
+  // - if either one of those is specified, and Crash dumps have not been explicitly disabled,
```
RealCLanger, line 3149: Shouldn't it say: "... and crash dumps have not explicitly been enabled," (not disabled)?

src/hotspot/share/runtime/globals.hpp
```diff
@@ -865,10 +865,16 @@ const intx ObjectAlignmentInBytes = 8;
           "JVM exits on the first occurrence of an out-of-memory error "    \
           "thrown from JVM")                                                \
                                                                             \
+  /* SapMachine 2021-05-21: Changed comment (we do not core on OOM) */      \
   product(bool, CrashOnOutOfMemoryError, false,                             \
-          "JVM aborts, producing an error log and core/mini dump, on the "  \
+          "JVM aborts, producing an error log, on the "                     \
```
RealCLanger, line 870: Should the possibility to explicitly enable core/mini dump and the printing of the thread stack be mentioned here?

src/hotspot/share/utilities/vmError.cpp
```diff
@@ -1834,3 +1834,13 @@ void VMError::controlled_crash(int how) {
   ShouldNotReachHere();
 }
 #endif // !ASSERT
+
+// SapMachine 2021-05-21: A wrapper for VMError::print_stack(..), public, for printing stacks
```
RealCLanger, line 1838: A wrapper for VMError::print_stack_trace (instead of VMError::print_stack), no?

src/hotspot/share/utilities/vmError.hpp
```diff
@@ -184,5 +184,9 @@ class VMError : public AllStatic {
   // Needed when printing signal handlers.
   NOT_WINDOWS(static const void* crash_handler_address;)
 
+  // SapMachine 2021-05-21: A wrapper for VMError::print_stack(..), public, for printing stacks
```
RealCLanger, line 187: A wrapper for VMError::print_stack_trace (instead of VMError::print_stack), no?

RealCLanger (changes requested):
Just a few remarks to the comments. Other than that, LGTM.

When we merge this, we should update https://github.com/SAP/SapMachine/wiki/Differences-between-SapMachine-and-OpenJDK.

Thanks for doing this!

tstuefe:
> Just a few remarks to the comments. Other than that, LGTM.
> 
> When we merge this, we should update https://github.com/SAP/SapMachine/wiki/Differences-between-SapMachine-and-OpenJDK.
> 
> Thanks for doing this!

Thank you Christoph!

I worked in your feedback. 

Cheers, Thomas

SapMachine:

SapMachine:

src/hotspot/share/prims/jvm.cpp
```diff
@@ -2913,6 +2913,11 @@ JVM_ENTRY(void, JVM_StartThread(JNIEnv* env, jobject jthread))
         JVMTI_RESOURCE_EXHAUSTED_OOM_ERROR | JVMTI_RESOURCE_EXHAUSTED_THREADS,
         os::native_thread_creation_failed_msg());
     }
+
+    // SapMachine 2021-05-21: All ...OnOutOfMemoryError switches should work for
+    //  thread creation failures too.
+    report_java_out_of_memory(os::native_thread_creation_failed_msg());
```
schmelter-sap, line 2919: In case of an OOM from the heap the report_java_out_of_memory() method is called before the JVMTI event is optionally generated. Is there a good reason to do this differently here?

src/hotspot/share/utilities/debug.cpp
```diff
@@ -361,10 +361,14 @@ void report_java_out_of_memory(const char* message) {
 
     if (CrashOnOutOfMemoryError) {
       tty->print_cr("Aborting due to java.lang.OutOfMemoryError: %s", message);
+      // SapMachine 2021-05-21: Print stack to stdout.
+      VMError::print_stack(tty);
```
schmelter-sap, line 365: I'm not sure if it is the best idea to do a heap dump for thread exhaustion too, since it can take a lot of time. And if the VM is killed during dumping of the heap we are missing the thread dump, which is probably more useful in this case, Plus it would be nice to have the thread dump even if we are not exiting the VM.

schmelter-sap (commented):
Some small remarks. Looks good otherwise.

src/hotspot/share/prims/jvm.cpp
tstuefe: No, as I wrote I was torn here. The question is whether we want things like CF jvmkill to have first dibs on OOM handling. Well maybe not :) I'll change this and do it before JVMTI.

src/hotspot/share/utilities/debug.cpp
tstuefe: Just to see if I understand, you propose:
- HeapDumpOnOutOfMemoryError should only be active for OOMs from Heap and Metaspace, whereas the other three switches should work for thread exhaustion too
- A stack should be written if any of these four switches is active? So, if -XX:OnOutOfMemoryError=myscript, we should print the stack too?

tstuefe:
> Some small remarks. Looks good otherwise.

Thanks Ralf. See Remarks inline.

SapMachine:

SapMachine:

tstuefe:
See also documentation of the new switches here:

https://github.com/SAP/SapMachine/wiki/Handling-of-XXXOnOutOfMemoryError-switches-in-the-SapMachine

please take a look as well.

src/hotspot/share/utilities/debug.cpp
```diff
@@ -349,8 +349,26 @@ void report_java_out_of_memory(const char* message) {
...
+    if ((HeapDumpOnOutOfMemoryError && !is_thread_exhaustion) ||
+        (OnOutOfMemoryError && OnOutOfMemoryError[0]) ||
+        CrashOnOutOfMemoryError || ExitOnOutOfMemoryError) {
+      // SapMachine 2021-05-21: Print stack to stdout
+      VMError::print_stack(tty);
+    }
+
     // create heap dump before OnOutOfMemoryError commands are executed
-    if (HeapDumpOnOutOfMemoryError) {
+    if (HeapDumpOnOutOfMemoryError && !is_thread_exhaustion) {
```
schmelter-sap, line 371: I think it would be best to always get the thread dump when we have thread exhaustion and it should be printed before we try a heap dump. Dumping the heap after dumping the threads is OK, since the most important information is already dumped.

tstuefe:
Okay, I changed this back to HeapDumpOnOutOfMemoryError handling thread exhaustion, but printing the thread stack before doing anything else.

I also added a test for CrashOnOutOfMemoryError in combination with ErrorFileToStdout to see limits on stdout.

SapMachine:

schmelter-sap (approved):
Looks good to me.

RealCLanger (dismissed):
Looks good to me now.

One remark/question, though: Isn't all core file writing disabled (by default) when setting -XX+CrashOnOutOfMemoryError? That means, not only when an OOM is the reason for the VM crash?

tstuefe:
> Looks good to me now.
> 
> One remark/question, though: Isn't all core file writing disabled (by default) when setting -XX+CrashOnOutOfMemoryError? That means, not only when an OOM is the reason for the VM crash?

Goddammit! :) You are right, good catch.

SapMachine:

tstuefe:
Okay I changed the logic so that cores are only switched off if we actually die from CrashOnOutOfMemoryError. Normal crashes still produce cores (tested).

RealCLanger:
> Okay I changed the logic so that cores are only switched off if we actually die from CrashOnOutOfMemoryError. Normal crashes still produce cores (tested).

Great, so now really LGTM 😄

RealCLanger (dismissed):

tstuefe:
Thanks Ralf & Christoph!

tstuefe:
@schmelter-sap can you approve pls?

schmelter-sap (approved):
Upps, I though I already approved it.

RealCLanger:
> Upps, I though I already approved it.

Yes, but it went stale 😄

SapMachine:

tstuefe:
@schmelter-sap @RealCLanger Can you approve again please? I had to modify the test to pass on Windows. The rest of the test errors seem unrelated. Thanks again!

RealCLanger (approved):

schmelter-sap (approved):
LGTM

PR #879, opened 2021-06-24
Merge to tag jdk-18+3

SapMachine:
please pull


ArnoZeller (approved):
Looks good to me.
The two issues on macOS are long standing issues - we should analyze them.

RealCLanger (approved):

PR #919, opened 2021-08-05
Merge to tag jdk-18+9

SapMachine:
please pull


chalstrick (commented):
For me the test results look ok. But the the test gc/metaspace/TestMetaspacePerfCounters.java#id3.TestMetaspacePerfCounters_id3
fails also on linux x86 64 . Can we ignore such a failing test

RealCLanger:
> For me the test results look ok. But the the test gc/metaspace/TestMetaspacePerfCounters.java#id3.TestMetaspacePerfCounters_id3
> fails also on linux x86 64 . Can we ignore such a failing test

Yes, the two test failures could be ignored in principal but the validate linux_x86_64 job is required so I restarted it.

RealCLanger (approved):

chalstrick (approved):

PR #951, opened 2021-09-07
950: Print out all flags in hs-err file

### LINKED ISSUE #950: Print out all flags in hs-err file
In hs-err file, print out all flags, not only the non-default flags.

Traditionally we have done this always, and it simplifies the logistics of error analysis since one does not have to cross-check VM switch default values in the sources.

SAP has always preferred expressive and extensive error reports at the cost of brevity; OpenJDK upstream does not, and therefore I do not believe I would be able to convince folks to do this change upstream. However, the change is super trivial, so it should not be an issue to carry this patch downstream only.

tstuefe:
Small change:
- print *all* flags, not just non-default flags, in the hs-err file
- but omit comments, they just bloat the hs-err file

Changes of doing this upstream are slim since Oracle prefers its hs-err file brief and succinct, while we don't care much for size but want to simplify error analysis as much as possible. It's a matter of taste.

fixes #950

### DIFF
--- a/src/hotspot/share/utilities/vmError.cpp
+++ b/src/hotspot/share/utilities/vmError.cpp
@@ -1008,11 +1008,14 @@ void VMError::report(outputStream* st, bool _verbose) {
   STEP("printing flags")
 
     if (_verbose) {
+      // SapMachine 2021-09-07:
+      // - print all values, not only default
+      // - comments are unnecessary bloat
       JVMFlag::printFlags(
         st,
-        true, // with comments
+        false, // with comments
         false, // no ranges
-        true); // skip defaults
+        false); // skip defaults
       st->cr();
     }
 


SapMachine:

RealCLanger (approved):
LGTM,

we should mention this behavioral difference here when it is integrated: https://github.com/SAP/SapMachine/wiki/Differences-between-SapMachine-and-OpenJDK

ArnoZeller (approved):

ArnoZeller:
Processbuilder test on windows failed with this error:
`[0.177s][warning][os,thread] Failed to start thread "Unknown thread" - _beginthreadex failed (EACCES) for attributes: stacksize: default, flags: CREATE_SUSPENDED STACK_SIZE_PARAM_IS.
`
 I guess we can ignore it. I am sure it is not related to this change.

PR #953, opened 2021-09-09
950: Print out all flags in hs-err file (17)

### LINKED ISSUE #950: Print out all flags in hs-err file

tstuefe:
(cherry picked from commit 6f0299770c357733d70cb28a73ae36a29d7907d5)

fixes #950

### DIFF
--- a/src/hotspot/share/utilities/vmError.cpp
+++ b/src/hotspot/share/utilities/vmError.cpp
@@ -1005,11 +1005,14 @@ void VMError::report(outputStream* st, bool _verbose) {
   STEP("printing flags")
 
     if (_verbose) {
+      // SapMachine 2021-09-07:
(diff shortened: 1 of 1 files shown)


RealCLanger (approved):

SapMachine:

ArnoZeller (approved):
Looks good - ship it.

PR #956, opened 2021-09-23
Merge to tag jdk-11.0.13+6

SapMachine:
please pull


RealCLanger (approved):
Looks good from my end.

On Linux aarch64 we see an intermittent error in hotspot/jtreg/compiler/aot/verification/vmflags/TrackedFlagTest.java which we can probably ignore.
The failing MacOS Version change is fixed with https://bugs.openjdk.java.net/browse/JDK-8269850 which I'm trying to get included into official OpenJDK 11.0.13.

We should merge this one.

ArnoZeller (approved):
Looks good regarding the comments of Christoph and the inclusion of the macOS fix.

PR #964, opened 2021-09-27
#963: Vitals: lock creation asserts after JDK-8273915 was merged

### LINKED ISSUE #963: Vitals: lock creation asserts after JDK-8273915 was merged
JDK-8273915 "Create 'nosafepoint' rank" changed mutex rankings. Since then, the Vitals crash on initialization in debug builds:

```
# To suppress the following error report, specify this argument                 
# after -XX: or in .hotspotrc:  SuppressErrorAt=/mutex.cpp:292                                          
#                                                                                              
# A fatal error has been detected by the Java Runtime Environment:                                                                                                
#                                                                                                                                                         
#  Internal Error (/shared/projects/openjdk/sapmachine/source/src/hotspot/share/runtime/mutex.cpp:292), pid=76528, tid=76533
#  assert(_rank <= nosafepoint || _safepoint_check_required == _safepoint_check_always) failed: Locks above nosafepoint rank should safepoint: Vitals Lock                       
#                                                                                                                                                                
# JRE version: OpenJDK Runtime Environment (18.0) (fastdebug build 18-internal+0-adhoc.thomas.source)                                                          
# Java VM: OpenJDK 64-Bit Server VM (fastdebug 18-internal+0-adhoc.thomas.source, mixed mode, tiered, compressed oops, compressed class ptrs, g1 gc, linux-amd64)
# Problematic frame:                                                                                                                                                                                  
# V  [libjvm.so+0x12ff4a7]  Mutex::Mutex(int, char const*, Mutex::SafepointCheckRequired, bool)+0xd7                                                                                                     
```

Vitals lock rank has to be adjusted to the new "nosafepoint"

tstuefe:
Oracle runtime groups rewrites Mutex rank checking etc. One of the changes was JDK-8273915 "Create 'nosafepoint' rank". Since then, the Vitals assert when initializing the Vitals lock.

For now, I changed the rank of the lock accordingly. I may have to rethink this further, but I believe this is fine for now.

Tests:
- runtime/Vitals jtreg tests locally
- manually tested that PrintVitalsAtExit works
- ... and that we still get Vitals when we crash.

fixes #963

### DIFF
--- a/src/hotspot/share/vitals/vitals.cpp
+++ b/src/hotspot/share/vitals/vitals.cpp
@@ -1193,7 +1193,7 @@ void sample_jvm_values(Sample* sample, bool avoid_locking) {
 
 bool initialize() {
 
-  g_vitals_lock = new Mutex(Mutex::leaf, "Vitals Lock", Mutex::_safepoint_check_never);
+  g_vitals_lock = new Mutex(Mutex::nosafepoint, "Vitals Lock", Mutex::_safepoint_check_never);
 
   if (!ColumnList::initialize()) {
     return false;


RealCLanger (approved):
I think this is great. Test error on linuxppc64le is something we see intermittently, nothing to do with this.

ArnoZeller (approved):
Looks good. The test failure on Linux PPC64le is an intermittent issue that is unrelated to the change.
Lets ship it!

PR #973, opened 2021-10-12
SapMachine #939: Add facility to trace uninstrumented malloc calls in VM and third party coding (17)

### LINKED ISSUE #939: Add facility to trace uninstrumented malloc calls in VM and third party coding
As internal prototypes showed, we can use glibc malloc hook in a safely manner to implement a malloc tracing facility which would work with uninstrumented code.

In short, we hook into glibc allocation. When allocation happens, we take a call stack (similar to how NMT does it), then file this callstack in an internal call site table.

tstuefe:
(cherry picked from commit bf45d7295c24e2583af047f2a8830a24ca2555ae)

Clean downport.

fixes #939


SapMachine:

PR #974, opened 2021-10-14
Merge to tag jdk-18+19

SapMachine:
please pull

PR #976, opened 2021-10-14
#975: Vitals: use native lock instead of hotspot Mutex

### LINKED ISSUE #975: Vitals: use native lock instead of hotspot Mutex
In the Vitals backend, we synchronize access to the shared sample table via a lock. This is needed since the table is accessed concurrently from two places:
- the sampler thread, when it adds a new sample
- the reporter, when it prints a report.

A report can be triggered from various places: jcmd, VM shutdown, error handling...

Today, we use a hotspot mutex for this synchronization, but since hotspot mutexes are changed upstream (and that work will continue for a while according to @coleenp) I'd like to use native mutexes instead to reduce merging and testing effort. We do not need the additional features hotspot mutexes give us (e.g. deadlock recognition).

tstuefe:
Hi,
can I have reviews for the following patch. It is supposed to solve current merge problems with jdk18-19 (https://github.com/SAP/SapMachine/pull/974) as well as prevent future problems, since hotspot mutexes will continue to change upstream.

Changes in detail:
- Added a new native lock implementation, super simple, for Posix and Windows.
  - Error handling is minimal: on Windows I use critical sections which have no error conditions, and on Posix I use a posix mutex where all possible error conditions are programming errors, hence the asserts.
- Removed the lock asserts since the new locks can't do this reliably and its not worth the complexity
- When printing the sample tables (which is one of the two lock regions) I got paranoid since the outputStream is unknown and may lock too (ttyLock etc). Therefore I first print to memory, then - outside the lock region - to the real stream. The memory is preallocated when Vitals are enabled, so no allocation happens when printing (which may be e.g. during error reporting).

The patch looks worse than it is due to indentation, use hide whitespace to minimize that.

Tests: 
- gtests & runtime/Vitals jtreg tests for debug, release, 32bit
- manual tests
- hotspot:tier1 is on the way.
- GHAs on the way.

fixes #975

### DIFF
--- a/src/hotspot/share/vitals/vitals.cpp
+++ b/src/hotspot/share/vitals/vitals.cpp
@@ -35,7 +35,6 @@
 #include "memory/metaspaceUtils.hpp"
 #include "memory/universe.hpp"
 #include "runtime/os.hpp"
-#include "runtime/mutex.hpp"
 #include "runtime/mutexLocker.hpp"
 #include "runtime/nonJavaThread.hpp"
 #include "runtime/thread.hpp"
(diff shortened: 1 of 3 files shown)


SapMachine:

PR #988, opened 2021-11-18
Merge to tag jdk-11.0.14+3

SapMachine:
please pull


RealCLanger:
Some test failures on windows and linux ppc64le, but I think no reason not to merge.

RealCLanger (approved):

chalstrick:
Hi Christoph, the validate-pr-11-windows_x86_64 tests show only one failure which I am not sure whether it's valid. But validate-pr-11-linux_ppc64le show more failures. Is this really acceptable?

RealCLanger:
> Hi Christoph, the validate-pr-11-windows_x86_64 tests show only one failure which I am not sure whether it's valid. But validate-pr-11-linux_ppc64le show more failures. Is this really acceptable?

The failures should at least not block integration, I think. But one should have a look some time. We have a new ppc64 node with a newer Ubuntu, so maybe that's the reason for at least some of the new issues.

chalstrick (approved):
lgtm

PR #995, opened 2021-12-03
Merge to tag jdk-18+26

SapMachine:
please pull


ArnoZeller (approved):
Looks good.
The issue for macOS is caused by a crashing compiler on the build machine. This is no showstopper for the SapMachine. We should just try to switch to a newer compiler.

RealCLanger:
> Looks good. The issue for macOS is caused by a crashing compiler on the build machine. This is no showstopper for the SapMachine. We should just try to switch to a newer compiler.

Thanks for the approval. I will, however, hold on to merging this until I have replaced the XCode in the build job.

RealCLanger (approved):

PR #1109, opened 2022-05-05
Merge to tag jdk-18.0.1.1+2

SapMachine:
please pull


RealCLanger (approved):
This is good. GHA errors occuring due to absence of JDK-8283017. This was only backported with 18.0.2. Test errors in Alpine build can be ignored, too as this is only beta.

chalstrick (approved):

Issue #1097, opened 2022-05-03
Vitals: April 22 Update


tstuefe:
Bring current updates to Vitals (April 22 version) to SapMachine.

PR #1120, opened 2022-05-14
SapMachine #1119: Vitals: Housekeeping May 22

### LINKED ISSUE #1119: Vitals: Housekeeping May 22
Some minor cleanup changes and fixes that came up during the development of #1112 which are best reviewed separately.

tstuefe:
These are some tiny issues and cleanups which came up during recent Vitals development and reviews.

Split into individual commits for easier review.

--------------

- [Fix bug where for jcmd VM.info the Vitals report settings were ignored and default was used instead](https://github.com/SAP/SapMachine/commit/20806d9eefaa879f2f402975f70eda0e317b8ce9)

Forgot to pass in the prepared options into the report function which causes the report function to use default values instead.

- [CPU times can overflow on long running machines](https://github.com/SAP/SapMachine/commit/f9622e589c1fb16e395f3731e05fdf5ecb89fcb2)

CPU times are read from /proc/stat/cpu and those numbers can get larger than 32-bit irregardless the bitness of the OS. Use 64-bit ints to prevent that.

- [Remove CPU-waiting column since it turns out to be unreliable](https://github.com/SAP/SapMachine/commit/515ea096d3d4e99e48e743129a1a373072dc97ae)

Cleanup change. Don't read iowait from /proc/stat/cpu since it makes not much sense with modern CPUs. The values were usually confusing, most of the time 0.

- [Remove BRK column; it was of questionable use and for that quite expensive to get](https://github.com/SAP/SapMachine/commit/a92bba37cfc13f338d44a75c7c91427b0319cb5f)

Cleanup change. I decided to remove the BRK column that shows the size of the traditional Unix heap. Even though this sometimes (with MALLOC_ARENA_MAX=1 and a bit of luck) was useful to determine the size of the C-heap, we have a better mechanism now (see the "cheap" columns introduced with #1097). And since the value is difficult to get (parsing the maps file) I got rid of it.

- [Remove code for older Linux kernels < 3.14](https://github.com/SAP/SapMachine/commit/a413d7f56be5f82c1935e67066ed39bf7ac2c959)

This removes coding needed for kernels < 3.14; not really needed anymore. 

--------------

fixes #1119

### DIFF
--- a/src/hotspot/os/linux/vitals_linux.cpp
+++ b/src/hotspot/os/linux/vitals_linux.cpp
@@ -96,55 +96,6 @@ public:
 
 };
 
-// Returns the sum of RSS and Swap for the process heap segment
-static value_t get_process_heap_size() {
-  FILE* f = ::fopen("/proc/self/smaps", "r");
-  value_t result = INVALID_VALUE;
-  if (f != NULL) {
-    char line[256];
-    int safety = 100000; // for safety reasons;
-    // Note: this does not guarantee atomicity wrt updates of the underlying file;
-    // however, we don't care here. We just weed out improbable values. Encountering
-    // inconsistencies due to concurrent updates should be very rare and probably manifest
-    // as unparseable lines.
-    int state = 0; // 1 - found segment, 2 - have rss 3 - have swap
-    long long rss = 0;
-    long long swap = 0;
-    while (state < 3 && ::fgets(line, sizeof(line), f) != NULL && safety > 0) {
-      // We look for something like this:
-      // "559f05393000-559f05671000 rw-p 00000000 00:00 0                          [heap]"
-      // ...
-      // RSS:     100 kB"
-      // ...
-      // Swap:    100 kB"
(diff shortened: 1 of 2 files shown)


SapMachine:

RealCLanger (approved):
Looks good, without thoroughly examining each change. 😉 

Test errors on Alpine can be ignored. On the MacOS platforms we have to check why all these sa tests are failing but this has for sure nothing to do with this PR.

ArnoZeller (approved):
Looks good.
Just want to mention that RHEL 6.10 still uses kernel 3.10 and has no MemAvailable.

tstuefe:
> Looks good. Just want to mention that RHEL 6.10 still uses kernel 3.10 and has no MemAvailable.

Thanks Arno! The only effect would be that this column will stay empty. I can live with that since it only affect a small minority of systems.

tstuefe:
Thank you @schmelter-sap @RealCLanger @ArnoZeller.

PR #1125, opened 2022-05-16
SapMachine #1124: Vitals: Add Container information

### LINKED ISSUE #1124: Vitals: June 22 package
Update Vitals to V3

tstuefe:
This patch adds cgroup information to the SapMachine Vitals.

We now print limits (memory, mem+swap, kernel mem) and usage (user space, kernel usage).

Note: we omit this information if the VM is assigned to the root cgroup memory controller. This is because in that case, no limits can be established, so printing out the cgroup info is kind of pointless and should closely resemble whatever is in `/proc/meminfo` anyway.

Also note that since most modern Linux distros use systemd, and systemd has its own memory controller, in almost all cases we will actually see the cgroup information.

Looks like this (in this example, docker was run with -m 8G, and it establishes a memory limit of 8g and - apparently - a mem+swap limit of 16g):

```
     cgroup-lim: cgroup memory limit
     cgroup-limsw: cgroup memory+swap limit
     cgroup-slim: cgroup memory soft limit
     cgroup-usg: cgroup memory usage
     cgroup-usgsw: cgroup memory+swap usage
     cgroup-kusg: cgroup kernel memory usage
```

```
------------cgroup------------
lim  limsw slim usg usgsw kusg
8.0g 16.0g      67m   67m   2m
8.0g 16.0g      67m   67m   2m
8.0g 16.0g      32m   32m   2m
8.0g 16.0g      19m   19m   2m
```

This PR also contains a large test addition. I added tests to better check the Vitals values for validity.

fixes #1124

### DIFF
--- a/src/hotspot/os/linux/vitals_linux.cpp
+++ b/src/hotspot/os/linux/vitals_linux.cpp
@@ -306,7 +306,7 @@ bool platform_columns_initialize() {
   // with /proc/meminfo and also root cgroup cannot have resource limits, so its less interesting.
   // In the future we also may omit if we run inside systemd, but I am not sure if systemd can limit
   // memory usage, and then those limits would be interesting
-  g_show_cgroup_info = !(context == ctx_unknown || context == ctx_root);
+  g_show_cgroup_info = (context != ctx_root);
 
   // System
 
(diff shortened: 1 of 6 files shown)


SapMachine:

SapMachine:

PR #1159, opened 2022-07-01
SapMachine #1124: Vitals: June 22 package (SapMachine 17)

tstuefe:
Downport of Vitals June 22 patch to SapMachine 17.

Minor things needed to be patched up, see 0a036dd0f30c69e70664a9962d160a11c2336f2c. 

- We have no "Platform.isMusl()" yet, so cherry-picked that from head.
- We have no "os::file_exists()" convenience wrapper yet, so cherry-picked that from head

The rest are small fish.

fixes #1124

PR #1171, opened 2022-07-05
SapMachine #1124: Vitals: June 22 package (SapMachine 18)

### LINKED ISSUE #1124: Vitals: June 22 package

tstuefe:
Backport to SapMachine 18. I originally thought I can leave this out since 19 is around the corner, but seems not.

As usual, the 18-specific changes you can find in the second commit: ea21757676876853c558a07290c03e74b442aeff

(in this case, I downported the patch from jdk19, not mainline, so the changes are minimal).

Changes:
- Platform.isMusl is missing, I re-added it.

fixes #1124

### DIFF
--- a/src/hotspot/os/linux/globals_linux.hpp
+++ b/src/hotspot/os/linux/globals_linux.hpp
@@ -82,6 +82,36 @@
           "resulting in file-backed shared mappings of the process to " \
           "be dumped into the corefile.")                               \
                                                                         \
+  /* SapMachine 2022-05-01: HiMemReport */                              \
+  product(bool, HiMemReport, false,                                     \
+         "VM writes a high memory report and optionally execute "       \
+         "additional jcmds if its rss+swap reaches 66%, 75% or 90% of " \
+         "a maximum. If the VM is containerized, that maximum is the "  \
+         "container memory limit at VM start. If the VM is not "        \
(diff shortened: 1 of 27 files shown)


SapMachine:

tstuefe:
Again with the weird GHA mac m1 build.  https://github.com/SAP/SapMachine/runs/7192055302?check_suite_focus=true . It really looks like we build without INCLUDE_NMT. Or with --disable-native-memory-tracking. Our own build works. We had this in the Vitals June 22 patch downport for SapMachine17 too (https://github.com/SAP/SapMachine/pull/1159). But interestingly enough, not in the downport for 11.

While the build error itself is a small issue indirectly caused by the same error that caused us not to be able minimal in 11 and 17, its indicative for something strange, since we never *should* build without NMT. NMT should be always on unless explicitly disabled on the configure command line. But the configure command line for the GHA mac m1 build looks fine.

I cannot reproduce this locally when building on a mac m1, and our Jenkins builds don't fail either, so I ignore it for now.

tstuefe:
All other test errors unrelated.

schmelter-sap (approved):
The 18 specific changes look good.

tstuefe:
Thanks @schmelter-sap and @RealCLanger !

PR #1243, opened 2022-08-25
SapMachine #1241: Vitals NMT "mlc" value can be too high

### LINKED ISSUE #1241: Vitals NMT "mlc" value can be too high
NMT "mlc" column can be significantly higher than RSS; which is strange. The error came in as a result of merging in https://bugs.openjdk.org/browse/JDK-8292072 and a subsequent Vitals-local fix done as part of the merge 20+11 https://github.com/SAP/SapMachine/pull/1231:

https://github.com/SAP/SapMachine/blob/4248b9c6e9d2b8d1be0880bbedbf80cc6837af27/src/hotspot/share/vitals/vitals.cpp#L1142-L1148

The problem is that we cannot use `MallocMemorySummary::as_snapshot();` directly since it won't make arena-vs-malloc-accounting corrections. This is a really messy workaround NMT has to do to get accounting for both arenas and mallocs right.

The right way is to use MemBaseline.

### LINKED ISSUE #1242: Write regression test for NMT mlc >>> RSS
See https://github.com/SAP/SapMachine/issues/1241. In tests, we allow NMT values to be larger than RSS since malloced memory does not have to be touched. We should rethink that.

tstuefe:
See issue text #1241 .

fixes #1241 
fixes #1242

### DIFF
--- a/src/hotspot/share/vitals/vitals.cpp
+++ b/src/hotspot/share/vitals/vitals.cpp
@@ -41,6 +41,7 @@
 #include "runtime/nonJavaThread.hpp"
 #include "runtime/thread.hpp"
 #include "runtime/threads.hpp"
+#include "services/memBaseline.hpp"
 #include "services/memTracker.hpp"
 #include "services/mallocTracker.hpp"
 #include "utilities/debug.hpp"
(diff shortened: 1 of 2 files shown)


SapMachine:

tstuefe:
Test errors unrelated

schmelter-sap (approved):
LGTM

RealCLanger (approved):

PR #1277, opened 2022-11-18
Merge to tag jdk-20+24

SapMachine:
please pull


RealCLanger (approved):
Failing tests on MacOS arm64 and Windows need some investigation but should not hold up this PR.

PR #1457, opened 2023-07-26
SapMachine (21) #1449: Store vitals samples for max/min values

### LINKED ISSUE #1449: Store vitals samples for max/min values
In addition to the short and long time tables, store a sample when specific values are at a maximum or minimum. This not only contains the information about the max/min value, but has the whole set of values at that time. So you can see for example the memory usage of the different types when the free memory was lowest or the committed memory the highest. For each value we store decide if we are interested in the max value, the min value or neither.

RealCLanger:
(cherry picked from commit ee6966d6266e6b8d0c3743f0bfb9c7393a695fae)

The description of this pull request goes here.

fixes #1449

### DIFF
--- a/src/hotspot/os/linux/vitals_linux.cpp
+++ b/src/hotspot/os/linux/vitals_linux.cpp
@@ -84,8 +84,8 @@ class CPUTimeColumn: public Column {
   }
 
 public:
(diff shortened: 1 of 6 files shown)


ArnoZeller (approved):
Looks reasonable.

schmelter-sap (approved):
Looks good

PR #1479, opened 2023-08-24
Merge to tag jdk-17.0.9+4

SapMachine:
please pull


MBaesken:
macos x86_64 : some timeout in a test, seems to be harmless

PR #1513, opened 2023-09-29
SapMachine #1512: ParallelGC: parallel scanning of large object arrays in old gen during young gc should be disabled by default

### LINKED ISSUE #1512: ParallelGC: parallel scanning of large object arrays in old gen during young gc should be disabled by default
Parallel scanning of large object arrays in the old generation should be disabled by default.

It should be possible to enable it with an -XX flag.

reinrich:
Introduce new vm option `UseParallelLargeArrayScanning` which controls parallel scanning of large object arrays in the old generation during young collections.

`UseParallelLargeArrayScanning` is disabled by default.

fixes #1512

### DIFF
--- a/src/hotspot/share/gc/shared/gc_globals.hpp
+++ b/src/hotspot/share/gc/shared/gc_globals.hpp
@@ -695,7 +695,7 @@
           constraint(GCCardSizeInBytesConstraintFunc,AtParse)               \
                                                                             \
   /* SapMachine 2023-09-25 */                                               \
-  product(bool, UseParallelLargeArrayScanning, false,                       \
+  product(bool, UseParallelLargeArrayScanning, false, EXPERIMENTAL,         \
           "Parallelize scanning of large object arrays in old gen "         \
           "when scanning roots for parallel young gc")
   // end of GC_FLAGS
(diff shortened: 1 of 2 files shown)


SapMachine:
Hello @reinrich, this pull request fulfills all formal requirements.

src/hotspot/share/gc/shared/gc_globals.hpp
```diff
@@ -692,7 +692,12 @@
   product(uint, GCCardSizeInBytes, 512,                                     \
           "Card table entry size (in bytes) for card based collectors")     \
           range(128, NOT_LP64(512) LP64_ONLY(1024))                         \
-          constraint(GCCardSizeInBytesConstraintFunc,AtParse)
+          constraint(GCCardSizeInBytesConstraintFunc,AtParse)               \
+                                                                            \
+  /* SapMachine 2023-09-25 */                                               \
+  product(bool, UseParallelLargeArrayScanning, false,                       \
```
TheRealMDoerr, line 698: Should it be EXPERIMENTAL? I guess we will remove the flag again once we have the final version backported.

reinrich: I didn't bother but probably it's better to make it experimental.

reinrich: Done :)

TheRealMDoerr (approved):

RealCLanger (approved):

PR #1516, opened 2023-10-04
SapMachine (11) #1512: ParallelGC: parallel scanning of large object arrays in old gen during young gc should be disabled by default

### LINKED ISSUE #1512: ParallelGC: parallel scanning of large object arrays in old gen during young gc should be disabled by default

RealCLanger:
Backport to SapMachine11.

fixes #1512


TheRealMDoerr (changes requested):
The flag specification doesn't work for 11u: You need to use "experimental(...)" instead of "product(...)".

RealCLanger:
> The flag specification doesn't work for 11u: You need to use "experimental(...)" instead of "product(...)".

OK, right. Fixed by force-push. Let's see now. :)

SapMachine:

TheRealMDoerr (approved):

reinrich (approved):

PR #1541, opened 2023-11-22
SapMachine #1540: Fix incorrect handling of invalid values in vitals extremas

### LINKED ISSUE #1540: Vital extremas doesn't handle invalid values correctly.
When vitals sample include invalid values, these are encoded as UINT64_MAX. This is handled by the extremas code as a maximum, but should be ignored instead.

schmelter-sap:
This fixes treating invalid values as extremas in the vitals code. Additionally we make sure to properly zero-terminate the buffer used for proc file reading, so invalid values are less likely.

fixes #1540

### DIFF
--- a/src/hotspot/os/linux/vitals_linux_oswrapper.cpp
+++ b/src/hotspot/os/linux/vitals_linux_oswrapper.cpp
@@ -84,8 +84,8 @@ public:
       return false;
     }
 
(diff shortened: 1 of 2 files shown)


RealCLanger (approved):
LGTM

ArnoZeller (approved):
LGTM

Issue #1550, opened 2023-11-28
Replace malloc trace relying on glibc malloc hooks


schmelter-sap:
Since the glibc malloc hooks are deprecated in 2.32 and removed in 2.34, the old malloc trace will not work anymore. It should be replaced by an implementation not relying on this glibc feature, so more platforms can use it.

PR #1551, opened 2023-11-28
SapMachine #1550: New malloc trace implementation

schmelter-sap:
This change implements a malloc trace without using the now deprecated glibc malloc hooks. This means the feature is now available on musl-based system and MacOsX too. In addition it can also track memory reclamation, so it can track the currently alive memory.

fixes #1550


SapMachine:

SapMachine:

MBaesken (commented):
Hi Ralf, some changed/added code parts still miss a comment tag 'SapMachine'  (e.g. CoreLibraries.gmk ) . Maybe better add it (it is done already at most places ) ?

Then there is some 'SapMachine RS' , not sure if the RS is intentionaly ? I think we usually just tag 'Sapmachine'  (but only an observation, maybe the 'old style' is okay too) ?

SapMachine:

make/modules/java.base/Copy.gmk
```diff
@@ -277,3 +277,18 @@ $(eval $(call SetupTextFileProcessing, CREATE_CLASSFILE_CONSTANTS_H, \
 TARGETS += $(CREATE_CLASSFILE_CONSTANTS_H)
 
 ################################################################################
+
```
RealCLanger, line 280: Please remove lines 280 and 281. Also add a # SapMachine annotation in shared code.

make/modules/java.base/lib/CoreLibraries.gmk
```diff
@@ -209,3 +209,21 @@ ifeq ($(call isTargetOs, aix), true)
   TARGETS += $(BUILD_LIBJLI_STATIC)
 
 endif
+
+##########################################################################################
```
RealCLanger, line 213: Add a # SapMachine annotation, instead of the ### line

make/test/JtregNativeHotspot.gmk
```diff
@@ -852,6 +852,9 @@ ifeq ($(call isTargetOs, linux), true)
   BUILD_HOTSPOT_JTREG_EXECUTABLES_LIBS_exeFPRegs := -ldl
   BUILD_HOTSPOT_JTREG_LIBRARIES_LIBS_libAsyncGetCallTraceTest := -ldl
   BUILD_HOTSPOT_JTREG_LIBRARIES_LDFLAGS_libfast-math := -ffast-math
+# SapMachine RS 2023-10-04
```
RealCLanger, line 855: Should rather be something like
`# SapMachine 2023-10-04 New malloc trace implementation`
So, without the initials and with some description

src/hotspot/os/linux/malloctrace/mallocTrace.cpp
```diff
@@ -29,6 +29,7 @@
 #include "malloctrace/assertHandling.hpp"
 #include "malloctrace/locker.hpp"
 #include "malloctrace/mallocTrace.hpp"
+#include "malloctrace/mallocTrace2.hpp"
```
RealCLanger, line 32: Please annotate

src/hotspot/os/linux/malloctrace/mallocTrace.cpp
```diff
@@ -137,6 +138,7 @@ class HookControl : public AllStatic {
 
   static void enable() {
     DEBUG_ONLY(verify();)
+    MallocStatistic::disable(NULL);
```
RealCLanger, line 141: Please annotate

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
RealCLanger, line 1: I suggest to name it mallocTracePosix.cpp

src/hotspot/os/posix/malloctrace/mallocTrace2.hpp
RealCLanger, line 1: Same here: mallocTracePosix.hpp

src/hotspot/share/utilities/debug.cpp
```diff
@@ -221,6 +226,13 @@ void report_fatal(VMErrorType error_type, const char* file, int line, const char
 
 void report_vm_out_of_memory(const char* file, int line, size_t size,
                              VMErrorType vm_err_type, const char* detail_fmt, ...) {
+  // SapMachine RS 2023-11-03: Check if we should to an emrgency dump for the malloc trace.
```
RealCLanger, line 229: Remove the initials from the annotation

src/hotspot/share/utilities/vmError.cpp
```diff
@@ -1667,6 +1671,11 @@ void VMError::report_and_die(int id, const char* message, const char* detail_fmt
                              Thread* thread, address pc, void* siginfo, void* context, const char* filename,
                              int lineno, size_t size)
 {
+#if defined(LINUX) || defined(__APPLE__)
+  // SapMachine RS 2023-09-18: Make sure we don't track allocations anymore.
```
RealCLanger, line 1675: Remove initials here

src/java.base/macosx/native/libjli/java_md_macosx.m
```diff
@@ -376,6 +376,32 @@ static void MacOSXStartup(int argc, char *argv[]) {
     }
     JLI_Snprintf(jvmcfg, so_jvmcfg, "%s%slib%sjvm.cfg",
                  jrepath, FILESEP, FILESEP);
+
+    /* SpaMachine RS 2023-09-18 */
```
RealCLanger, line 380: Remove initials and add some comment like "new malloc trace"

src/java.base/share/native/libjli/java.c
```diff
@@ -850,6 +850,71 @@ CheckJvmType(int *pargc, char ***argv, jboolean speculative) {
     return jvmtype;
 }
 
+// SapMachine RS 2023-09-18
```
RealCLanger, line 853: Same here, -initials +comment

src/java.base/share/native/libjli/java.h
```diff
@@ -172,6 +172,9 @@ void AddOption(char *str, void *info);
 jboolean IsWhiteSpaceOption(const char* name);
 jlong CurrentTimeMicros();
 
+// SapMachine RS 2023-09-18
```
RealCLanger, line 175: -initials + comment

src/java.base/unix/native/libjli/java_md.c
```diff
@@ -346,6 +346,36 @@ CreateExecutionEnvironment(int *pargc, char ***pargv,
     mustsetenv = RequiresSetenv(jvmpath);
     JLI_TraceLauncher("mustsetenv: %s\n", mustsetenv ? "TRUE" : "FALSE");
 
+    /* SapMachine RS 2023-09-18 */
```
RealCLanger, line 349: -intials + comment

src/java.base/unix/native/libmallochooks/mallochooks.c
```diff
@@ -0,0 +1,1016 @@
+/*
+ * Copyright (c) 2023 SAP SE. All rights reserved.
+ * Copyright (c) 2023, Oracle and/or its affiliates. All rights reserved.
```
RealCLanger, line 3: Don't need to credit Oracle here (and in other new files that you added and that are not substantially copied from something from OpenJDK)

src/java.base/unix/native/libmallochooks/mallochooks.h
```diff
@@ -0,0 +1,92 @@
+/*
+ * Copyright (c) 2023 SAP SE. All rights reserved.
+ * Copyright (c) 2023, Oracle and/or its affiliates. All rights reserved.
```
RealCLanger, line 3: Don't add Oracle

test/hotspot/jtreg/runtime/malloctrace/MallocHooksTest.java
```diff
@@ -0,0 +1,702 @@
+/*
+ * Copyright (c) 2023 SAP SE. All rights reserved.
+ * Copyright (c) 2023, Oracle and/or its affiliates. All rights reserved.
```
RealCLanger, line 3: Remove Oracle

test/hotspot/jtreg/runtime/malloctrace/exetestmallochooks.c
```diff
@@ -0,0 +1,248 @@
+/*
+ * Copyright (c) 2023 SAP SE. All rights reserved.
+ * Copyright (c) 2023, Oracle and/or its affiliates. All rights reserved.
```
RealCLanger, line 3: Remove Oracle

test/hotspot/jtreg/runtime/malloctrace/libfakemallochooks.c
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright (c) 2023 SAP SE. All rights reserved.
+ * Copyright (c) 2023, Oracle and/or its affiliates. All rights reserved.
```
RealCLanger, line 3: Remove Oracle

test/hotspot/jtreg/runtime/malloctrace/libtestmallochooks.c
```diff
@@ -0,0 +1,267 @@
+/*
+ * Copyright (c) 2023 SAP SE. All rights reserved.
+ * Copyright (c) 2023, Oracle and/or its affiliates. All rights reserved.
```
RealCLanger, line 3: Remove Oracle

RealCLanger (changes requested):
Did a first pass and have some formal requests

make/modules/java.base/Copy.gmk
```diff
@@ -277,3 +277,18 @@ $(eval $(call SetupTextFileProcessing, CREATE_CLASSFILE_CONSTANTS_H, \
...
+################################################################################
+# Copy mallochook.h
+
+ifeq ($(call isTargetOs, windows), false)
```
MBaesken, line 284: This tests for 'not windows', but in the next file you test for 'is Linux or macOS'.  What about AIX we have this now too in Sapmachine ?  I think the makefile checks should be consistent (I guess test for 'is Linux or macOS') .

make/modules/java.base/lib/CoreLibraries.gmk
```diff
@@ -209,3 +209,21 @@ ifeq ($(call isTargetOs, aix), true)
...
+
+ifeq ($(call isTargetOs, linux macosx), true)
+  LIBJLI_SRC_DIRS := $(call FindSrcDirsForComponent, java.base, libmallochooks)
+
+  $(eval $(call SetupJdkLibrary, BUILD_MALLOCHOOKS, \
```
MBaesken: Usually (~90%) those library-targets in OpenJDK are named BUILD_LIBxyz (e.g. BUILD_LIBNET, BUILD_LIBPREFS etc.) Why not do it this way here too ? If you rename it, is the `SRC := `setting still needed?

make/modules/java.base/lib/CoreLibraries.gmk
```diff
@@ -209,3 +209,21 @@ ifeq ($(call isTargetOs, aix), true)
...
+      NAME := mallochooks, \
+      SRC := $(TOPDIR)/src/java.base/unix/native/libmallochooks, \
+      OUTPUT_DIR := $(INSTALL_LIBRARIES_HERE), \
+      OPTIMIZATION := LOW, \
```
MBaesken, line 205: Is there a good reason to go for LOW optimization? Would it still work with higher opt-settings?

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2706 @@
...
+#endif
+
+// To test in jtreg tests use
+// JTREG="JAVA_OPTIONS=-XX:+UseMallocHooks -XX:+MallocTraceAtStartup -XX:+MallocTraceDump -XX:MallocTraceDumpInterval=10 -XX:MallocTraceDumpOutput=`pwd`/mtrace_@pid.txt -XX:ErrorFile=`pwd`/hs_err%p.log"
+
+// A simple smoke test
+// jconsole -J-XX:+UseMallocHooks -J-XX:+MallocTraceAtStartup -J-XX:+MallocTraceDump -J-XX:MallocTraceStackDepth=12 -J-XX:MallocTraceDumpInterval=10
+
+
+// Some compile time constants for the maps.
```
MBaesken: Regarding all the compile time constants below - would the coding still work with adjusted settings for those values (as long as they follow the rules e.g. power of two etc.) ?  Have you considered  to make some of them adjustable at runtime ?

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2706 @@
...
+  buf[end - start] = '\0';
+
+  char* found_end;
+  uint64_t result = (uint64_t) strtoull(buf, &found_end, 10);
+
+  if ((found_end != end) && (*found_end != '\0')) {
+    *error = "Could not parse integer";
+  }
+
+  return (uint64_t) atoll(buf);
```
MBaesken: Is there a good reason to use strtoll  AND atoll ? The few places in OpenJDK codebase using strtoll  just use this without atoll.

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2706 @@
...
+          *error = "Unexpected character";
+          return 0;
+        }
+    }
+
+    pos++;
+  }
+
+  if (pos != start) {
+    *error = "time wihtout unit";
```
MBaesken: without not wihtout

src/hotspot/os/linux/malloctrace/mallocTrace.cpp
schmelter-sap: Is this needed in SapMachine-only code?

src/hotspot/os/linux/malloctrace/mallocTrace.cpp
schmelter-sap: As above this is SapMachine-only code.

RealCLanger: Ah, you're right.

src/hotspot/os/linux/malloctrace/mallocTrace.cpp
RealCLanger: you're right. Forget this comment then. 😄

make/modules/java.base/Copy.gmk
schmelter-sap: Done

make/modules/java.base/lib/CoreLibraries.gmk
schmelter-sap: Done

SapMachine:

SapMachine:

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2705 @@
...
+  int get_slot(address to_check);
+
+public:
+  AddressHashSet(real_funcs_t* funcs, bool enabled);
+  ~AddressHashSet();
+
+  bool contains(address to_check);
+  bool add(address to_add);
+  size_t allocated();
+  double load();
```
MBaesken: what is the double about that `'load()' ` returns ? Maybe add a short comment line.

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2705 @@
...
+  return 1.0 * _count / (_mask + 1);
+}
+
+// A pthread mutex usable in arrays.
+struct Lock {
+  char            _pre_pad[DEFAULT_CACHE_LINE_SIZE];
+  pthread_mutex_t _lock;
+};
+
+class Locker : public StackObj {
```
MBaesken: Did you consider reusing something form Hotspot , there seems to be PlatformMutex class available.  Or do you have special needs that the existing stuff is not really fulfilling?

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2705 @@
...
+  }
+}
+
+
+//
+//
+//
+//
+// Class StatEntry
+//
```
MBaesken: Could you please add some comments here what it is about?   Or just delete the large comment, it does not help much as it is now.

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2705 @@
...
+  uint64_t   _count;
+  address    _frames[1];
+
+public:
+  StatEntry(size_t hash, size_t size, int nr_of_frames, address* frames) :
+    _next(NULL),
+    _hash_and_nr_of_frames((hash * (MAX_FRAMES + 1)) + nr_of_frames),
+    _size(size),
+    _count(1) {
+    memcpy(_frames, frames, sizeof(address) * nr_of_frames);
```
MBaesken: Should there be a check that not too much is attempted to copy into `_frames` ?

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2705 @@
...
+    if ((register_hooks == NULL) || (get_real_funcs == NULL)) {
+      if (UseMallocHooks) {
+        st->print_raw_cr("Could not find preloaded libmallochooks while -XX:+UseMallocHooks is set. " \
+                         "This usually happens if the VM is not loaded via the JDK launcher (e.g. " \
+                         "java.exe). In this case you must preload the library by setting the " \
+                         "following environment variable: ");
+        print_needed_preload_env(st);
+      } else {
+        st->print_cr("Could not find preloaded libmallochooks. Try using -XX:+UseMallocHooks " \
+                     "Vm option to automatically preload it using the JDK launcher. Or you can set " \
```
MBaesken: Maybe better  use 'VM option' not 'Vm option' ?

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2705 @@
...
+
+volatile registered_hooks_t*
+                  MallocStatisticImpl::_rainy_day_hooks_ptr = &MallocStatisticImpl::_rainy_day_hooks;
+
+#define CAPTURE_STACK() \
+  address frames[MAX_FRAMES + FRAMES_TO_SKIP]; \
+  uint64_t ticks = _detailed_stats ? Ticks::now().nanoseconds() : 0; \
+  int nr_of_frames = 0; \
+  if (_max_frames <= 2) { \
+    /* Skip, since we will fill it it later anyway. */ \
```
MBaesken: Probably you mean 'fill it in' ?

make/modules/java.base/Copy.gmk
schmelter-sap: Initially I included AIX as well, but removed it since. I've modified the test to Linux and MacOSX

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
schmelter-sap: I will do it at the end. Not sure what happens to people reviewing this currently if I do it now.

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2705 @@
...
+
+  if ((result != NULL) && should_track(hash) && malloc_not_suspended()) {
+    CAPTURE_STACK();
+
+    if (_track_free) {
+      record_allocation(result, hash, nr_of_frames, frames);
+    } else {
+      // Here we track the really allocated size, since it might be very different
+      // from the requested one.
+      record_allocation_size(real_malloc_size(result), nr_of_frames, frames);
```
MBaesken: looks like the whole block is repeated in the similar methods below again; could  we move it into a helper function  or some kind of macro to shorten the code and avoid some repetition ?

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2705 @@
...
+      ss.print_cr("Same hash " UINT64_FORMAT " for %p and %p", (uint64_t) hash, ptr, entry->ptr());
+      ss.print_raw_cr("Current stack:");
+
+      for (int i = 0; i < nr_of_frames; ++i) {
+        ss.print("  [" PTR_FORMAT "]  ", p2i(frames[i]));
+        os::print_function_and_library_name(&ss, frames[i], tmp, sizeof(tmp), true, true, false);
+        ss.cr();
+      }
+
+      ss.print_raw_cr("Orig stack:");
```
MBaesken: Maybe better write 'Original'  (if this is meant) ?

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2705 @@
...
+      return false;
+    }
+
+    _alloc_maps_alloc[i] = new (mem) Allocator(sizeof(AllocEntry), 2048, _funcs);
+    _alloc_maps_mask[i] = ALLOC_MAP_INIT_SIZE - 1;
+    _alloc_maps_size[i]._val = 0;
+    _alloc_maps_limit[i] = (int) ((_alloc_maps_mask[i] + 1) * MAX_ALLOC_MAP_LOAD);
+    _alloc_maps[i] = (AllocEntry**) _funcs->calloc(_alloc_maps_mask[i] + 1, sizeof(AllocEntry*));
+
+    if (mem == NULL) {
```
MBaesken: You check already for `mem == NULL`  a couple of lines above. Why again, is there some hidden assignment to mem again?

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2705 @@
...
+  uint64_t k = 1024;
+  double perc = 0.0;
+  if (total > 0) {
+    perc = 100.0 * mem / total;
+  }
+
+  if ((int64_t) mem < 0) {
+    mem = -((int64_t) mem);
+    st->print("*neg* ");
+  }
```
MBaesken: When would a negative value happen? What does it mean?

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2705 @@
...
+}
+
+MallocTraceDumpDCmd::MallocTraceDumpDCmd(outputStream* output, bool heap) :
+  DCmdWithParser(output, heap),
+  _dump_file("-dump-file", "If given the dump command writes the result to the given file. " \
+             "Note that the filename is interpreted by the target VM. You can use " \
+             "'stdout' or 'stderr' as filenames to dump via stdout or stderr of " \
+             "the target VM", "STRING", false),
+  _filter("-filter", "If given we only print a stack if it contains a function matching " \
```
MBaesken: Maybe give an example or tell more in detail what 'matching' means here ?  Substring found? regex ? Something else ? Would be helpful for the user to get a bit more info.

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2705 @@
...
+          "the given string.", "STRING", false),
+  _max_entries("-max-entries", "The maximum number of entries to dump.", "INT", false, "10"),
+  _dump_fraction("-fraction", "If > 0 we dunp the given fraction of allocated bytes " \
```
MBaesken: dump not dunp ; and is the value meant between in percent (between 0 and 100) or between 0 and 1 ? Maybe give an example, always helpful for the user (0.5 or 50 for 50% I guess).

src/hotspot/share/runtime/globals.hpp
```diff
@@ -549,6 +549,101 @@ const int ObjectAlignmentInBytes = 8;
           "When DumpVitalsAtExit is set, the file name prefix for the "     \
           "output files (default is sapmachine_vitals_<pid>).")             \
                                                                             \
+  /* SapMachine 2023-09-18: malloc trace */                                 \
+  product(bool, UseMallocHooks, false,                                      \
+          "Preloads the malloc hooks library needed for the malloc trace. " \
+          "This flag only works when using a JDK launcher. Otherwise the "  \
+          "library has to preloaded by hand.")                              \
```
MBaesken: has to **be** preloaded ?

src/hotspot/share/runtime/globals.hpp
```diff
@@ -549,6 +549,101 @@ const int ObjectAlignmentInBytes = 8;
...
+          "print an error message and exit. Otherwise the error is "        \
+          "silently ignored.")                                              \
+                                                                            \
+  product(bool, MallocTraceTrackFree, true,                                 \
+          "If set the malloc trace enabled at startup also tracks "         \
+          "deallocation of memory")                                         \
+                                                                            \
+  product(ccstr, MallocTraceEnableDelay, "0s",                              \
+          "If > 0 seconds and MallocTraceAtStartup is enabled, we delay "   \
+          "the startup by the given amount of time. Can use s, m, h or d "  \
```
MBaesken: Is this about the startup of tracing/tracking? If so, better add the info here, like ', we delay the startup of ... by' .

src/hotspot/share/runtime/globals.hpp
```diff
@@ -549,6 +549,101 @@ const int ObjectAlignmentInBytes = 8;
...
+                                                                            \
+  product(bool, MallocTraceUseBacktrace, true,                              \
+          "If set we use the backtrace() call to samples the stacks of "    \
+          "the malloc trace enabled at startup. Note that while this "      \
+          "creates better stack traces, it is also slower and not "         \
+          "supported on every system. If not supported, this option is "    \
+          "simply ignored.")                                                \
+                                                                            \
+  product(ccstr, MallocTraceUnwindLibName, "libunwind.so.8",                \
+          "The name of the libunwind to load.")                             \
```
MBaesken: Maybe tell the user what it is good for? Can we use a full path? Or Just the name of the lib?

src/hotspot/share/runtime/globals.hpp
```diff
@@ -549,6 +549,101 @@ const int ObjectAlignmentInBytes = 8;
...
+  product(bool, MallocTraceDetailedStats, false,                            \
+          "If enabled we collect more detailed statistics for the malloc "  \
+          "trace enabled at startup. This costs some performance.")         \
+                                                                            \
+  product(bool, MallocTraceDumpOnError, false,                              \
+          "If enabled and the malloc trace is enabled too we do an "        \
+          "emergency dump on native out-of-memory errors.")                 \
+                                                                            \
+  product(uintx, MallocTraceRainyDayFund, 1* M,                             \
+          "The size of the rainy day fund to use when doing an emergency "  \
```
MBaesken, line 605: Is the 'rainy day fund' naming/meaning explained?  I guess it is similar to SAP JVM rainy day fund?

src/hotspot/share/runtime/globals.hpp
```diff
@@ -549,6 +549,101 @@ const int ObjectAlignmentInBytes = 8;
...
+          "If > 0 it limits the number of entries printed. If no sort "     \
+          "is specified via -XX:MallocTraceDumpSort, we sort by "           \
+          "size.")                                                          \
+                                                                            \
+  product(uintx, MallocTraceDumpFraction, 0,                                \
+          "If > 0 we dunp the given fraction of allocated bytes "           \
+          "(or allocated objects if sorted by count). In that case the "    \
+          "-XX:MallocTraceDumpMaxEntries option is ignored.")               \
+                                                                            \
+  product(bool, MallocTraceDumpHideDumpAlllocs, true,                       \
```
MBaesken: Alllocs - really 3 l's ?

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2705 @@
...
+
+static void dump_from_flags(bool on_error) {
+  DumpSpec spec;
+  char const* file = MallocTraceDumpOutput;
+  spec._on_error = on_error;
+  spec._filter = MallocTraceDumpFilter;
+  spec._sort_by_count = MallocTraceDumpSortByCount;
+  spec._max_entries = MallocTraceDumpMaxEntries;
+  spec._dump_fraction = MallocTraceDumpFraction;
+  spec._hide_dump_allocs = MallocTraceDumpHideDumpAlllocs;
```
MBaesken: Really Alllocs with 3 l's ?

src/java.base/unix/native/libmallochooks/mallochooks.c
```diff
@@ -0,0 +1,1015 @@
...
+#define FREE_REPLACEMENT     __libc_free
+#define MEMALIGN_REPLACEMENT __libc_memalign
+#define VALLOC_REPLACEMENT   __libc_valloc
+#define PVALLOC_REPLACEMENT  __libc_pvalloc
+#endif
+
+#elif defined(_AIX)
+
+#else
+  // This must be musl. Since they are cool they don't set a define.
```
MBaesken: :-)

src/java.base/unix/native/libmallochooks/mallochooks.c
```diff
@@ -0,0 +1,1015 @@
...
+#if !defined(MALLOC_REPLACEMENT)
+  print_size(fallback_buffer_pos - fallback_buffer);
+  print(" bytes used for fallback\n");
+  print_size(fallback_buffer_end - fallback_buffer_pos);
+  print(" bytes not used for fallback\n");
+#endif
+
+#endif
+
+  // Now the rest can be assigned, since even the fallack methods are not
```
MBaesken: fallback not fallack

src/hotspot/share/utilities/debug.cpp
schmelter-sap: Fixed

src/java.base/share/native/libjli/java.h
schmelter-sap: Fixed

SapMachine:

SapMachine:

reinrich (changes requested):
Hi Ralf,
I think I've looked at most of the changes now. Time to post what I've collected so far.
Great job in any case :)
Cheers, Richard.

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2675 @@
...
+
+  for (int i = 0; i < NR_OF_ALLOC_MAPS; ++i) {
+    if (pthread_mutex_init(&_alloc_maps_lock[i]._lock, NULL) != 0) {
+      fatal("Could not initialize lock");
+    }
+  }
+}
+
+bool MallocStatisticImpl::enable(outputStream* st, TraceSpec const& spec) {
+  initialize();
```
reinrich, line 1565: Done already at startup.

make/modules/java.base/lib/CoreLibraries.gmk
```diff
@@ -209,3 +209,21 @@ ifeq ($(call isTargetOs, aix), true)
...
+# SapMachine 2023-11-28: build libmallochooks
+
+ifeq ($(call isTargetOs, linux macosx), true)
+  LIBJLI_SRC_DIRS := $(call FindSrcDirsForComponent, java.base, libmallochooks)
+
+  $(eval $(call SetupJdkLibrary, BUILD_MALLOCHOOKS, \
+      NAME := mallochooks, \
+      SRC := $(TOPDIR)/src/java.base/unix/native/libmallochooks, \
+      OUTPUT_DIR := $(INSTALL_LIBRARIES_HERE), \
+      OPTIMIZATION := LOW, \
```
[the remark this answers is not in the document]
schmelter-sap: To keep the size of the .so file small. And there seems to be not much to optimize anyways, since most of the work is just forwarding the calls.

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2675 @@
...
+
+#if !defined(__APPLE__)
+#include <malloc.h>
+#endif
+
+// To test in jtreg tests use
+// JTREG="JAVA_OPTIONS=-XX:+UseMallocHooks -XX:+MallocTraceAtStartup -XX:+MallocTraceDump -XX:MallocTraceDumpInterval=10 -XX:MallocTraceDumpOutput=`pwd`/mtrace_@pid.txt -XX:ErrorFile=`pwd`/hs_err%p.log"
+
+// A simple smoke test
+// jconsole -J-XX:+UseMallocHooks -J-XX:+MallocTraceAtStartup -J-XX:+MallocTraceDump -J-XX:MallocTraceStackDepth=12 -J-XX:MallocTraceDumpInterval=10
```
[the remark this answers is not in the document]
schmelter-sap: I've forgotten to adjust these when I changed the options for automatic dumps. This should be fixed with the latest commit.

src/hotspot/os/linux/malloctrace/mallocTrace.cpp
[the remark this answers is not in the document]
reinrich: > The whole mallocTrace.cpp file is SapMachine-only.

Yeah, writing "This is SapMachine only code" I meant of course the whole file and not the single include line :)
And I do think the copyright year should be updated. Sorry for being pedantic 🤓

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2675 @@
...
+void MallocStatisticImpl::cleanup() {
+  for (int i = 0; i < NR_OF_STACK_MAPS; ++i) {
+    cleanup_for_stack_map(i);
+  }
+
+  for (int i = 0; i < NR_OF_ALLOC_MAPS; ++i) {
+    cleanup_for_alloc_map(i);
+  }
+
+  if (_funcs != NULL) {
```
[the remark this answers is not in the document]
schmelter-sap: That would crash if _funcs was not initialized, which can happen. Having _rainy_day_fund being NULL is OK, since free handles this aynway.

make/modules/java.base/lib/CoreLibraries.gmk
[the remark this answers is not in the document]
reinrich: `LOW` translates to -O2 and `HIGH` to -O3. `LOW` is good enough.

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2675 @@
...
+  initialize();
+  Locker lock(_malloc_stat_lock);
+
+  if (_enabled) {
+    if (spec._force) {
+      _enabled = false;
+      setup_hooks(NULL, st);
+      cleanup();
+
+      st->print_raw_cr("Disabling already running trace first.");
```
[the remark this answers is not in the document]
schmelter-sap: FIxed.

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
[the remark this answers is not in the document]
reinrich: I see. You need to check `_funcs != NULL` in `cleanup_for_stack_map` and `cleanup_for_alloc_map` or just return from `cleanup` if `_funcs` is not available.

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2675 @@
...
+volatile uint64_t MallocStatisticImpl::_tracked_ptrs;
+volatile uint64_t MallocStatisticImpl::_not_tracked_ptrs;
+volatile uint64_t MallocStatisticImpl::_failed_frees;
+void*             MallocStatisticImpl::_rainy_day_fund;
+Lock              MallocStatisticImpl::_rainy_day_fund_lock;
+
+volatile registered_hooks_t*
+                  MallocStatisticImpl::_rainy_day_hooks_ptr = &MallocStatisticImpl::_rainy_day_hooks;
+
+#define CAPTURE_STACK() \
```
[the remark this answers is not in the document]
schmelter-sap: That worked.

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2675 @@
...
+  uint64_t hash = (uint64_t) ptr;
+  hash = (~hash) + (hash << 21);
+  hash = hash ^ (hash >> 24);
+  hash = (hash + (hash << 3)) + (hash << 8);
+  hash = hash ^ (hash >> 14);
+  hash = (hash + (hash << 2)) + (hash << 4);
+  hash = hash ^ (hash >> 28);
+  hash = hash + (hash << 31);
+
+  assert(hash == ptr_hash_backup(ptr), "Must be the same");
```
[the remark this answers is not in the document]
schmelter-sap: I think I better remove the backup function. It is really the same, just replaced shifts and adds by direct multiplication.

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2675 @@
...
+    for (int i = 1; i < 63; ++i) {
+      uint64_t pow = ((uint64_t) 1) << i;
+      _to_track_limit = pow / only_nth;
+      double diff = pow / (double) _to_track_limit - only_nth;
+
+      _to_track_mask = pow - 1;
+
+      if (diff > -0.1 && diff < 0.1) {
+        break;
+      }
```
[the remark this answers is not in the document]
schmelter-sap: That's what I've done initially. But this version allows to approximate non-power-of-2 values too and the check itself is nearly as fast (masking and comparing with threshold).

src/hotspot/share/runtime/globals.hpp
```diff
@@ -552,6 +552,101 @@ const int ObjectAlignmentInBytes = 8;
           "When DumpVitalsAtExit is set, the file name prefix for the "     \
           "output files (default is sapmachine_vitals_<pid>).")             \
                                                                             \
+  /* SapMachine 2023-09-18: malloc trace */                                 \
```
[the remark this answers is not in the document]
schmelter-sap: I don't see a value in making them diagnostic. It's just makes the use more annoying.

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2675 @@
...
+    }
+
+    size_t entry_size = sizeof(StatEntry) + sizeof(address) * (_max_frames - 1);
+    _stack_maps_alloc[i] = new (mem) Allocator(entry_size, 256, _funcs);
+    _stack_maps_mask[i] = STACK_MAP_INIT_SIZE - 1;
+    _stack_maps_size[i]._val = 0;
+    _stack_maps_limit[i] = (int) ((_stack_maps_mask[i] + 1) * MAX_STACK_MAP_LOAD);
+    _stack_maps[i] = (StatEntry**) _funcs->calloc(_stack_maps_mask[i] + 1, sizeof(StatEntry*));
+
+    if (mem == NULL) {
```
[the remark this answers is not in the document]
schmelter-sap: Good catch. Should be fixed in the latest commit.

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2675 @@
...
+      st->print_raw_cr("Could not allocate the alloc map!");
+      cleanup();
+
+      return false;
+    }
+  }
+
+  if (_funcs == NULL) {
```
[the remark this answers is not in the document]
schmelter-sap: If we cannot get all the real malloc functions, we exit during the initialization of the malloc hooks lib anyway. So I think the check is not needed at all.

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2675 @@
...
+      }
+
+      st->print_raw_cr("VM arguments:");
+      Arguments::print_summary_on(st);
+
+      return false;
+    }
+  }
+
+  _funcs = get_real_funcs();
```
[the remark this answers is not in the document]
schmelter-sap: get_real_funcs() will never return NULL, since we exit if we cannot get one of the real malloc functions at startup.

src/java.base/unix/native/libmallochooks/mallochooks.c
```diff
@@ -0,0 +1,1015 @@
...
+EXPORT registered_hooks_t* malloc_hooks_active_hooks() {
+  if (registered_hooks == &empty_registered_hooks) {
+    return NULL;
+  }
+
+  return (registered_hooks_t*) registered_hooks;
+}
+
+EXPORT real_funcs_t* malloc_hooks_get_real_funcs() {
+  return &real_funcs;
```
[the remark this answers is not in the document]
schmelter-sap: We exit the process if this happens before this function could be called.

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2675 @@
...
+      Atomic::add(&_tracked_ptrs, (uint64_t) 1);
+    } else {
+      Atomic::add(&_not_tracked_ptrs, (uint64_t) 1);
+    }
+  }
+
+  return (hash & _to_track_mask) < _to_track_limit;
+}
+
+#define malloc_not_suspended() (pthread_getspecific(_malloc_suspended) == NULL)
```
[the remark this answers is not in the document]
schmelter-sap: The impact is small, since it just checks a thread-relative field for the pthread_getspecific() call.

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2675 @@
...
+
+// Must be a power of two minus 1.
+#define MAX_FRAMES 31
+
+// The number of top frames to skip.
+#define FRAMES_TO_SKIP 0
+
+// Must be a power of two.
+#define NR_OF_STACK_MAPS 16
+#define NR_OF_ALLOC_MAPS 16
```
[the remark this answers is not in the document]
schmelter-sap: Would this work in SapMachine17 too?

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2675 @@
...
+//
+//
+
+class StatEntry {
+private:
+  StatEntry* _next;
+  uint64_t   _hash_and_nr_of_frames;
+  uint64_t   _size;
+  uint64_t   _count;
+  address    _frames[1];
```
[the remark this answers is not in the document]
schmelter-sap: Never saw this feature, although it seems not to be standard C++ (https://en.wikipedia.org/wiki/Flexible_array_member).

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2675 @@
...
+    to_check = to_check->next();
+  }
+
+  // Need a new entry. Fail silently if we don't get the memory.
+  void* mem = _stack_maps_alloc[idx]->allocate();
+
+  if (mem != NULL) {
+    StatEntry* entry = new (mem) StatEntry(hash, to_add, nr_of_frames, frames);
+    assert(hash == entry->hash(), "Must be the same");
+    assert(nr_of_frames == entry->nr_of_frames(), "Must be equal");
```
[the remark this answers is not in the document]
schmelter-sap: Good  Idea.

reinrich (commented):
I didn't notice that GH buffered my replies because I was collecting more review comments. I'm flushing them now.

src/hotspot/os/posix/malloctrace/mallocTrace2.hpp
```diff
@@ -0,0 +1,212 @@
...
+    _dump_file(NULL),
+    _filter(NULL),
+    _max_entries(0),
+    _hide_dump_allocs(true),
+    _on_error(false),
+    _sort_by_count(false),
+    _dump_fraction(100),
+    _internal_stats(false) {
+  }
+};
```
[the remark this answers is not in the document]
schmelter-sap: The header file should contain the interface to handle the trace, so I think if enabling and printing methods are available, the definition of the specs should be in the header file, since otherwise you cannot call them.

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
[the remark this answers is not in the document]
schmelter-sap: Yes, the loop is really not needed (expect if smaller masks are faster to use ;)). Why 42 and not 62?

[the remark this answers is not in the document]
reinrich: 42 (actually 43) is sufficient to even trace every 1,000,000th malloc with an error < 0.1 (and it is such a special number too).
There's a dependency to the maximal hash value. You could introduce a  constant for it and derive `pow` from it.
62, 42, maximal hash, what ever you like better. It would be nice though to get rid of the loop.

SapMachine:

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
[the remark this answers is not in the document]
schmelter-sap: Ok, I've added a static bool field as a first check, since it move all the code which changes the key into a lock protected region, so we  don't have problem with concurrent modifications of the field.

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2675 @@
...
+      if (should_track(hash) && malloc_not_suspended()) {
+        record_allocation(result, hash, nr_of_frames, frames);
+      }
+    } else if ((old_size < size) && should_track(hash) && malloc_not_suspended()) {
+      // Track the additional allocate bytes. This is somewhat wrong, since
+      // we don't know the requested size of the original allocation and
+      // old_size might be greater.
+      record_allocation_size(size - old_size, nr_of_frames, frames);
+    }
+  }
```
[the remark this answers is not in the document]
schmelter-sap: In the free() and some parts of realloc() we actually need to track even if malloc_not_suspended() returns false.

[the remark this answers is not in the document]
reinrich: > In the free() and some parts of realloc() we actually need to track even if malloc_not_suspended() returns false.

The original lines I'm referring to do not track anything if malloc_not_suspended() returns false. Only the stack is captured then. I thought my suggestion is equivalent but simpler. Is it not?

SapMachine:

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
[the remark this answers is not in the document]
reinrich: Please note that I suggested to hoist `should_track(hash) && malloc_not_suspended()` up to the if-statement in L998 of version 53f23816129784f7364c545a5a75e61858744397, L1009 with [current version](https://github.com/SAP/SapMachine/commit/00108968c38894f7e161a0eeb72294244b51eb18).
It's confusing how Github displays the suggestion. I've done the change locally and git highlighted the change beautifully.
I'll add a new comment on the most recent version.

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2672 @@
...
+      if (should_track(hash) && !malloc_suspended()) {
+        record_allocation(result, hash, nr_of_frames, frames);
+      }
+    } else if ((old_size < size) && should_track(hash) && !malloc_suspended()) {
+      // Track the additional allocate bytes. This is somewhat wrong, since
+      // we don't know the requested size of the original allocation and
+      // old_size might be greater.
+      record_allocation_size(size - old_size, nr_of_frames, frames);
+    }
+  }
```
reinrich, line 1023: You can hoist the condition `should_track(hash) && !malloc_suspended()` up to the if statement in L1009
```
@@ -1006,15 +1006,13 @@ void* MallocStatisticImpl::realloc_hook(void* ptr, size_t size, void* caller_add
 
   uint64_t hash = ptr_hash(result);
 
-  if ((result != NULL) && should_track(hash)) {
+  if ((result != NULL) && should_track(hash) && !malloc_suspended()) {
     address frames[MAX_FRAMES + FRAMES_TO_SKIP];
     int nr_of_frames = capture_stack(frames, (address) realloc, (address) caller_address);
 
     if (_track_free) {
-      if (should_track(hash) && !malloc_suspended()) {
-        record_allocation(result, hash, nr_of_frames, frames);
-      }
-    } else if ((old_size < size) && should_track(hash) && !malloc_suspended()) {
+      record_allocation(result, hash, nr_of_frames, frames);
+    } else if (old_size < size) {
       // Track the additional allocate bytes. This is somewhat wrong, since
       // we don't know the requested size of the original allocation and
       // old_size might be greater.
```

SapMachine:

SapMachine:

SapMachine:

SapMachine:

SapMachine:

SapMachine:

reinrich (commented):
A few more comments but we're almost there.

SapMachine:

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2675 @@
...
+
+#else
+
+  AllocEntry(uint64_t hash, StatEntry* entry, AllocEntry* next) :
+    _hash(hash),
+    _entry(entry),
+    _next(next) {
+  }
+
+#endif
```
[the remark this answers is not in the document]
reinrich: Ping, have you seen this one?

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2675 @@
...
+  int idx = (int) (hash & (NR_OF_ALLOC_MAPS - 1));
+  Locker locker(_alloc_maps_lock[idx]);
+
+  if (!_enabled) {
+    return;
+  }
+
+  int slot = (hash / NR_OF_ALLOC_MAPS) & _alloc_maps_mask[idx];
+
+  // Should not already be in the table, so we remove the check in the optimized version
```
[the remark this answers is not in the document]
reinrich: Ping, have you seen this one?

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2675 @@
...
+
+    return entry;
+  }
+
+  return NULL;
+}
+
+void MallocStatisticImpl::record_allocation(void* ptr, uint64_t hash, int nr_of_frames, address* frames) {
+  assert(_track_free, "Only used for detailed tracking");
+  size_t size = _funcs->malloc_size(ptr);
```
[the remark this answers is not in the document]
reinrich: Ping, have you seen this one?

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2675 @@
...
+  assert(_track_free, "Only used for detailed tracking");
+
+  int idx = (int) (hash & (NR_OF_ALLOC_MAPS - 1));
+  Locker locker(_alloc_maps_lock[idx]);
+
+  if (!_enabled) {
+    return NULL;
+  }
+
+  int slot = (hash / NR_OF_ALLOC_MAPS) & _alloc_maps_mask[idx];
```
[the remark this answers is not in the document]
reinrich: Ping? I'd suggest to introduce `HashMapData::get_slot(uint64_t hash)`

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2675 @@
...
+    }
+  }
+
+  _funcs = get_real_funcs();
+  register_hooks(hooks);
+
+  return true;
+}
+
+uint64_t MallocStatisticImpl::ptr_hash(void* ptr) {
```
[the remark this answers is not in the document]
reinrich: Ping?

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2675 @@
...
+          *error = "Unexpected character";
+          return 0;
+        }
+    }
+
+    pos++;
+  }
+
+  if (pos != start) {
+    *error = "time wihtout unit";
```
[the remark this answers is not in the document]
reinrich: Ping?

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2675 @@
...
+    } else {
+      char const* pid_tag = strstr(file, "@pid");
+
+      if (pid_tag != NULL) {
+        size_t len = strlen(file);
+        size_t first = pid_tag - file;
+        char buf[32768];
+        jio_snprintf(buf, sizeof(buf), "%.*s" UINTX_FORMAT "%s",
+                    first, file, os::current_process_id(), pid_tag + 4);
+        fileStream fs(buf, "at");
```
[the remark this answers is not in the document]
reinrich: Ping?

src/hotspot/os/posix/malloctrace/mallocTrace2.cpp
```diff
@@ -0,0 +1,2675 @@
...
+          // in reality.
+        } else {
+          entries[idx] = (StatEntryCopy*) result;
+        }
+
+        nr_of_entries[idx] = max_entries;
+      }
+    } else {
+      nr_of_entries[idx] = 0;
+      failed_alloc = true;
```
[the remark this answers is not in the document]
reinrich: Ping?

reinrich:
Hi @schmelter-sap, I've marked some older comments as resolved. There are still a bunch open that might be worth addressing though.

SapMachine:

SapMachine:

SapMachine:

SapMachine:

SapMachine:

SapMachine:

SapMachine:

SapMachine:

SapMachine:

SapMachine:

SapMachine:

MBaesken:
> [src/hotspot/share/runtime/globals.hpp](https://github.com/SAP/SapMachine/pull/1551/files#diff-521dc8e4560864f444a550fc314476e4695d5fd55aebb71fdb30e9d16aa1589c)

Still little typo :  "library has to preloaded by hand."   -> I think you mean has to be preloaded

MBaesken:
product(bool, MallocTraceDumpHideDumpAlllocs, true,

Really 3 lll  in Alllocs this looks very strange.

MBaesken (approved):

src/hotspot/os/posix/malloctrace/mallocTracePosix.cpp
```diff
@@ -0,0 +1,2652 @@
...
+
+  for (int i = 0; i < max_entries; ++i) {
+    int max_pos = -1;
+    StatEntryCopy* max = NULL;
+
+    // Find the largest entry not currently printed.
+    if (spec._sort_by_count) {
+      for (int j = 0; j < NR_OF_STACK_MAPS; ++j) {
+        if (curr_pos[j] < nr_of_entries[j]) {
+          if ((max == NULL) || (max->_count < entries[j][curr_pos[j]]._count)) {
```
reinrich: I think you need to guard against `entries[j] == NULL`

src/hotspot/os/posix/malloctrace/mallocTracePosix.cpp
```diff
@@ -0,0 +1,2652 @@
...
+            max = &entries[j][curr_pos[j]];
+            max_pos = j;
+          }
+        }
+      }
+    } else {
+      for (int j = 0; j < NR_OF_STACK_MAPS; ++j) {
+        if (curr_pos[j] < nr_of_entries[j]) {
+          if ((max == NULL) || (max->_size < entries[j][curr_pos[j]]._size)) {
```
reinrich: I think you need to guard against `entries[j] == NULL`

src/java.base/unix/native/libmallochooks/mallochooks.c
```diff
@@ -0,0 +1,583 @@
...
+
+  errno = errno_backup;
+}
+
+static void print_error(char const* msg) {
+  write_safe(2, msg, strlen(msg));
+}
+
+
+static void unepected_call() {
```
reinrich:
```suggestion
static void unexpected_call() {
```

reinrich (changes requested):
Just 3 more nits...
BTW: don't you want to use `nullptr` instead of `NULL`?

SapMachine:

SapMachine:

schmelter-sap:
> Just 3 more nits... BTW: don't you want to use `nullptr` instead of `NULL`?

Since I want to backport to sapjvm where nullptr doesn't work I would avoid it for now.

src/hotspot/os/posix/malloctrace/mallocTracePosix.cpp
reinrich: I see now that the dereference of `entries[j]` is not reached because `nr_of_entries[j] == 0` if `entries[j] == NULL`

src/hotspot/os/posix/malloctrace/mallocTracePosix.cpp
reinrich: I see now that the dereference of `entries[j]` is not reached because `nr_of_entries[j] == 0` if `entries[j] == NULL`.
I wonder if static code analysis is smart enough ;)

SapMachine:

reinrich (approved):
Looks good!
It reassures me to know that we have such a good tool for analyzing native memory issues now.
Cheers, Richard.

SapMachine:

src/hotspot/os/linux/malloctrace/mallocTrace.cpp
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 SAP SE. All rights reserved.
+ * Copyright (c) 2024 SAP SE. All rights reserved.
```
RealCLanger, line 2: Should be 2021, 2024

RealCLanger (approved):
Looks good, I made a minor comment regarding one license year update.

When merging we need to carefully squash and clean up the commit comment. I could do this, if you are ok with that.

RealCLanger:
And it seems a MallocHooks test is failing on MacOS and linux ppc64le now. Can you please have a look and fix before we merge this?

SapMachine:

RealCLanger:
Test runtime/malloctrace/MallocHooksTest.java still seems to fail on Mac. Probably a locale issue?

SapMachine:

SapMachine:

PR #1570, opened 2023-12-13
Merge to tag jdk-23+1

SapMachine:
please pull


RealCLanger:
GHA failure on x86 will be fixed with [JDK-8321542](https://bugs.openjdk.org/browse/JDK-8321542).
Then we see an error in TestStubAllocFailure.java which is being fixed under [JDK-8321400](https://bugs.openjdk.org/browse/JDK-8321400). On AIX there's also a recurring timeout in TestStringEncodingJumbo, but I guess we can look at that later.

TheRealMDoerr (approved):
TestStubAllocFailure problems are known and not critical (test bug).

PR #1584, opened 2024-01-10
SapMachine #1583: Include more info in exception used in vitals tests

### LINKED ISSUE #1583: Include more information when pattern matching fails in vitals tests
Include the output against which the patterns are matched when matching fails.

schmelter-sap:
We need to see the output which doesn't include the expected message.

fixes #1583


TheRealMDoerr (approved):
LGTM.

PR #1594, opened 2024-01-29
SapMachine #1593: Implement a lean alternative for jcmd

### LINKED ISSUE #1593: Implement a lean alternative for jcmd
We often see usage of jcmd in (cloud) containers to get information needed for support. Since these environments tend to be configured with tight resources, it is important that the tool itself doesn't use much memory. Otherwise the container could be killed for exceeding its allowed memory usage. Since jcmd starts a whole JVM it would be nice to have a small native program which does mostly the same. One comfort feature which could be missing is specifying the connected VM by anything apart from a pid.

schmelter-sap:
This is aa simple C-base implementation of jcmd. Apart from only taking a pid, it offers the same features, but uses much less memory (and starts vastly faster).

Here is what a jcmd has to do:
- Check if /tmp/.java_pid\<pid> exists.
- If not create /tmp/.attach_pid\<pid> and send a SIGQUIT to the given pid. This starts the attach thread in the VM, which then creates the /tmp/.java_pid\<pid> socket for communication.
- Connect to the /tmp/.java_pid\<pid> file socket and write a command consisting of some scaffolding and the parameters to jcmd.
- Write the received output from the file socket to stdout (omitting the result code returned as the first line).

fixes #1593

### DIFF
--- a/make/modules/java.base/Launcher.gmk
+++ b/make/modules/java.base/Launcher.gmk
@@ -87,4 +87,17 @@ ifeq ($(call isTargetOs, macosx aix linux), true)
   TARGETS += $(BUILD_JSPAWNHELPER)
 endif
 
+# SapMachine 2024-01-24: Build a lean version of jcmd on Linux
+ifeq ($(call isTargetOs, linux), true)
+  $(eval $(call SetupJdkExecutable, BUILD_LEANJCMD, \
+      NAME := leanjcmd, \
+      SRC := $(TOPDIR)/src/$(MODULE)/unix/native/leanjcmd, \
+      OPTIMIZATION := LOW, \
+      CFLAGS := $(CFLAGS_JDKEXE) -I$(TOPDIR)/src/$(MODULE)/unix/native/libjava, \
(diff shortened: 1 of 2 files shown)


SapMachine:

MBaesken:
Maybe add a simple  '-help'  output to the leanjcmd to educate the user about the difference to the 'good' old Java based jcmd?

RealCLanger:
Could you add a smoketest for leanjcmd?

Also, is this already packaged into JDK/JRE images or would that need additional modifications?

simonis:
Looks like you want to take a look at the jattach utility which does exactly what you want (and a little more :)

https://github.com/jattach/jattach

RealCLanger:
> Looks like you want to take a look at the jattach utility which does exactly what you want (and a little more :)
> 
> https://github.com/jattach/jattach

Yep, we are aware of it. However, we'd probably like to have a tool which is always there in SapMachine.

SapMachine:

schmelter-sap:
Since we now have aprof available, this is not needed anymore.

PR #1650, opened 2024-04-23
SapMachine (22) #1646: Provide a mechanism to augment the Http KeepAliveCache key

### LINKED ISSUE #1646: Provide a mechanism to augment the Http KeepAliveCache key
A customer is accessing an HttpServer backend via HttpURLConnection through JAX-WS. Connections get authenticated by some special mechanism on a per-socket base and subsequent HTTP requests need to happen on the authenticated connection. So the matching authenticated connection has to be retrieved from the cache. To achieve this, we add a field of type ThreadLocal<String> to sun.net.www.http.KeepAliveCache that can be armed with a value and serves as an additional part of the KeepAliveCache key.

We have thought about alternative solutions for the customer but at the moment there seem to be none. The feature does not affect the API and is only enabled when system property `com.sap.jvm.UseHttpKeepAliveCacheKeyExtension` is set to `true`. To use this feature, client code needs an export statement like `--add-exports java.base/sun.net.www.http=ALL-UNNAMED`.

RealCLanger:
(cherry picked from commit 78be7ddfa35787bbac51fb5ee1e8df58c370ac5b)

fixes #1646

### DIFF
--- a/src/java.base/share/classes/sun/net/www/http/KeepAliveCache.java
+++ b/src/java.base/share/classes/sun/net/www/http/KeepAliveCache.java
@@ -81,6 +81,9 @@ public class KeepAliveCache
         userKeepAliveProxy = getUserKeepAliveSeconds("proxy");
     }
 
+    // SapMachine 2024-04-12: Provide additional key field for KeepAliveCache entries (for FRUN)
+    public static final ThreadLocal<String> connectionID = new ThreadLocal<>();
+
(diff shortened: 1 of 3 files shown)


SapMachine:

schmelter-sap (approved):
Looks good to me.

TheRealMDoerr (approved):

PR #1652, opened 2024-04-23
Fix test of new malloc trace

schmelter-sap:
This fixes a race condition which might leads to spurious test failures.

fixes #1550


SapMachine:

RealCLanger:
What exactly is the race condition that you fix here? (I'm not so deep in the code that it would be obvious to me... 😉)

RealCLanger (approved):

MBaesken (approved):

PR #1698, opened 2024-06-19
SapMachine #1670: Improve patch for jcmd GC.heap_dump modification

### LINKED ISSUE #1670: jcmd GC.heap_dump without options should write to location given by -XX:HeapDumpPath, if set
Currently jcmd command GC.heap_dump only works with an additionally provided file name.
Syntax : GC.heap_dump [options] <filename>

In case the JVM has the XX-flag HeapDumpPath set, we should support an additional mode where the <filename> is optional.
In case the filename is NOT set, we take the HeapDumpPath (file or directory).

new syntax :
GC.heap_dump [options] <filename> ... has precedence over second option
GC.heap_dump [options] … in case -XX:HeapDumpPath=p is set

This would be a simplification e.g. for support cases where a filename or directory is set at JVM startup with -XX:HeapDumpPath=p and writing to the path is intended/recommended for usage also in the jcmd case.

See also https://bugs.openjdk.org/browse/JDK-8327769

RealCLanger:
The SapMachine [commit to modify the jcmd GC.heap_dump](ab72d4e43a708764b3e6a6173e9f2999cf4bc17d) was taken directly from the OpenJDK proposal. In that form, however, it creates a too large upstream diff which could be reduced.

fixes #1670

### DIFF
--- a/src/hotspot/share/services/diagnosticCommand.cpp
+++ b/src/hotspot/share/services/diagnosticCommand.cpp
@@ -492,10 +492,10 @@ void FinalizerInfoDCmd::execute(DCmdSource source, TRAPS) {
 }
 
 #if INCLUDE_SERVICES // Heap dumping/inspection supported
-// SapMachine 2024-05-10: HeapDumpPath for jcmd
 HeapDumpDCmd::HeapDumpDCmd(outputStream* output, bool heap) :
                            DCmdWithParser(output, heap),
-  _filename("filename","Name of the dump file", "STRING", false, "if no filename was specified, but -XX:HeapDumpPath=hdp is set, path hdp is taken"),
+  // SapMachine 2024-05-10: HeapDumpPath for jcmd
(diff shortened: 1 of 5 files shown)


SapMachine:

MBaesken (changes requested):
I think we miss (now?) a couple of // Sapmachine  comments (e.g. heapDumper.cpp).
The globals.hpp change was omitted now; this is okay with me if it is seen better this way (to keep diffs low).

RealCLanger:
> I think we miss (now?) a couple of // Sapmachine comments (e.g. heapDumper.cpp).

Yes, it was my plan to add them after everything builds and tests are green 😄 

> The globals.hpp change was omitted now; this is okay with me if it is seen better this way (to keep diffs low).

No, it is already in the sapmachine code and won't be rolled back. The actual fix after these modifications can only be seen by diffing all files against openjdk.

SapMachine:

SapMachine:

SapMachine:

MBaesken (approved):

schmelter-sap (approved):
The diff now looks small compared to the version without MBs changes.

PR #1715, opened 2024-06-23
SapMachine #1714: Add fallback for early malloc for musl-based systems

### LINKED ISSUE #1714: MallocHooksTest fails on newer versions of Alpine Linux
On newer versions of Alpine Linux, the MallocHooksTest fails. This is caused by a call to malloc, when the real malloc functions could not yet be determined via dlsym. An earlier version of the malloc hooks had fallback code for this case, but since it was never triggered in practice and made the code ugly, it was removed. 

Link to the failing tests:
https://test.sapmachine.only.sap:8443/servlet/testresult?testid=258228791&show_similar_failed_tests=true

schmelter-sap:
This handles the case on musl-based systems, where malloc is called too early and we have not determined the rewal malloc functions. 

fixes #1714

### DIFF
--- a/src/java.base/unix/native/libmallochooks/mallochooks.c
+++ b/src/java.base/unix/native/libmallochooks/mallochooks.c
@@ -121,6 +121,8 @@ static void* calloc_by_malloc(size_t elems, size_t size);
 static int posix_memalign_by_aligned_alloc(void** ptr, size_t align, size_t size);
 static void* memalign_by_aligned_alloc(size_t align, size_t size);
 
+#define NEEDS_MALLOC_FALLBACK
+
 static real_malloc_funcs_t impl = {
   (malloc_func_t*) LOAD_DYNAMIC,
   calloc_by_malloc,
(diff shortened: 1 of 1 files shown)


SapMachine:

src/java.base/unix/native/libmallochooks/mallochooks.c
```diff
@@ -121,6 +121,8 @@ static void* calloc_by_malloc(size_t elems, size_t size);
 static int posix_memalign_by_aligned_alloc(void** ptr, size_t align, size_t size);
 static void* memalign_by_aligned_alloc(size_t align, size_t size);
 
+#define NEEDS_MALLOC_FALLBACK
```
RealCLanger, line 124: How does this define make sense? Can't you just add the new code unconditionally? Or is there also some other place that could do `#undef NEEDS_MALLOC_FALLBACK` ?

schmelter-sap: Currently it's only defined for musl-based systems.

RealCLanger: Ah, you're right. I should have looked where exactly the define resides... 😄

RealCLanger (approved):

ArnoZeller (approved):
LGTM

PR #1716, opened 2024-06-25
SapMachine (23) #1714: Add fallback for early malloc for musl-based systems

### LINKED ISSUE #1714: MallocHooksTest fails on newer versions of Alpine Linux

RealCLanger:
clean cherry-pick from commit 1106762a444e97830a99f025e650dea919e4aa8b

fixes #1714


schmelter-sap (approved):
LGTM

ArnoZeller (approved):
LGTM

SapMachine:

Issue #1693, opened 2024-06-12
Support starting child processes in an own process group


RealCLanger:
Add a SapMachine internal API to allow processes started through ProcessBuilder::start to be leader of an own process group instead of belonging to the process group of the parent VM.

Also provide a method to terminate process groups.

A minimal solution for this has already been implemented in SapMachine 11 through #203.

PR #1730, opened 2024-07-02
SapMachine #1693: Add support for starting child processes in an own process group also for Windows

RealCLanger:
This also adds Windows support for process groups.

fixes #1693


SapMachine:

src/java.base/windows/native/libjava/ProcessImpl_md.c
```diff
@@ -343,6 +349,16 @@ static jlong processCreate(
                     {
                         win32Error(env, L"CreateProcess");
                     } else {
+                        /* SapMachine 2024-07-01: process group extension */
+                        if (processGroup) {
+                            __declspec(align(8)) HANDLE hJob = CreateJobObject(NULL, NULL);
+                            if (!hJob) {
+                                win32Error(env, L"CreateJobObject");
+                            }
+                            (*env)->SetLongArrayRegion(env, processGroup, 0, 1, (jlong *) &hJob);
```
schmelter-sap: win32Error will throw an exception, so you should not call SetLongArrayRegion with a pending exception. And in that case we should not call AssignProcessToJobObject.

RealCLanger: ok, I guess youre right. Fixed.

SapMachine:

src/java.base/windows/native/libjava/ProcessImpl_md.c
```diff
@@ -477,6 +498,15 @@ Java_java_lang_ProcessImpl_closeHandle(JNIEnv *env, jclass ignored, jlong handle
     return (jboolean) CloseHandle((HANDLE) handle);
 }
 
+/* SapMachine 2024-07-01: process group extension */
+JNIEXPORT void JNICALL
+Java_java_lang_ProcessImpl_terminateProcessGroup(JNIEnv *env, jclass ignored, jlong hJob)
+{
+    if (TerminateJobObject((HANDLE) hJob, 1) == 0){
```
schmelter-sap: Missing space between ')' and '{'. And is there any reason to use 1 as the exit value (which usually is regarded as an error).

RealCLanger: The space is fixed. I think the exit code 1, usually an error code, is fine because when a process is killed from outside I guess it can rather be regarded as non-normal exit.

SapMachine:

schmelter-sap (approved):
Looks good.

src/java.base/share/classes/jdk/internal/access/JavaLangProcessAccess.java
ansteiner, line 1: master should be renamed, as detected by "In Solidarity"

ansteiner (approved):
Think about renaming master as considered by "In Solidarity"

src/java.base/share/classes/jdk/internal/access/JavaLangProcessAccess.java
RealCLanger: Good catch. Thanks.

SapMachine:

PR #1745, opened 2024-07-10
SapMachine (11) #1550: Implement new malloc trace for SapMachine 11

schmelter-sap:
This is mostly taken from the SAPJVM implementation, which is also based on JDK11. 

fixes #1550


SapMachine:

RealCLanger (commented):
What about this [hunk](https://github.com/SAP/SapMachine/blob/471cf1704b2e4f41fc5ffecdfc99526be5497d8e/src/hotspot/os/linux/malloctrace/mallocTrace.cpp#L134) which exists in sapmachine17. Why don't we need it in 11?

It also isn't guarded by #ifdefs in sapmachine21 and above.

Also, the build on linuxx86 failed.

SapMachine:

SapMachine:

SapMachine:

RealCLanger (approved):

src/java.base/unix/native/libmallochooks/mallochooks.c
```diff
@@ -0,0 +1,523 @@
...
+  print("Resolving '");
+  print(symbol);
+  print("'\n");
+
+  *dest = dlsym(RTLD_NEXT, symbol);
+
+  if (*dest == NULL) {
+    print_error(symbol);
+    print_error(" not found!\n");
+    /* Don't exit if we don't find the method. It is enough to fail
```
RealCLanger, line 158: This is a diff to upstream. Why?

schmelter-sap: Because we support a glibc version on x86_64 which doesn't have an aligned_alloc function. So if you run on such an old version, we don't want to fail since the aligned_alloc function will not be called anyway. But on newer versions we still want to be able to track it.

RealCLanger: Alright then.

reinrich (approved):
Lgtm.

PR #1748, opened 2024-07-11
SapMachine #1693: Fix initialization of JavaLangProcessAccess

RealCLanger:
Initialization of the Shared Secret JavaLangProcessAccess goes through class java.lang.Process but the static initializers were in java.lang.ProcessImpl.

I also sneaked in a little copyright fix.

fixes #1693

### DIFF
--- a/src/java.base/unix/classes/java/lang/ProcessImpl.java
+++ b/src/java.base/unix/classes/java/lang/ProcessImpl.java
@@ -574,7 +574,7 @@ final class ProcessImpl extends Process {
     // SapMachine 2024-07-01: process group extension
     private static native int terminateProcessGroup(long pid, boolean force);
 
-    protected void terminateProcessGroup(boolean force) throws IOException {
+    void terminateProcessGroup(boolean force) throws IOException {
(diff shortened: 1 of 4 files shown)


SapMachine:

src/java.base/unix/classes/java/lang/ProcessImpl.java
```diff
@@ -576,7 +574,7 @@ public String toString() {
     // SapMachine 2024-07-01: process group extension
     private static native int terminateProcessGroup(long pid, boolean force);
 
-    private void terminateProcessGroup(boolean force) throws IOException {
+    protected void terminateProcessGroup(boolean force) throws IOException {
```
schmelter-sap: Do you really need to make it protected? I would have thought making it package private would be enough.

RealCLanger: Good catch. Fixed.

SapMachine:

ansteiner (approved):

MBaesken (approved):

Issue #1502, opened 2023-09-15
Prepare SapMachine 21


RealCLanger:
Cherry-pick some changes for the release of SapMachine 21

PR #1756, opened 2024-07-15
SapMachine (17): Cherry-pick 8329109: Threads::print_on() tries to print CPU time for terminated GC threads

RealCLanger:
We want to cherry-pick the backport of [JDK-8329109](https://bugs.openjdk.org/browse/JDK-8329109) for the July CPU. Due to SapMachine diffs, the upstream patch did not apply cleanly, I had to massage it.

fixes #1502

### DIFF
--- a/src/hotspot/share/runtime/thread.cpp
+++ b/src/hotspot/share/runtime/thread.cpp
@@ -3589,7 +3589,6 @@ void Threads::add(JavaThread* p, bool force_daemon) {
   p->set_on_thread_list();
 
   _number_of_threads++;
-
   oop threadObj = p->threadObj();
   bool daemon = true;
   // Bootstrapping problem: threadObj can be null for initial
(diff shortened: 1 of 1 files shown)


SapMachine:

ansteiner (approved):

src/hotspot/share/runtime/thread.cpp
```diff
@@ -3863,13 +3863,7 @@ void Threads::print_on(outputStream* st, bool print_stacks,
...
-  cl.do_thread(VMThread::vm_thread());
-  Universe::heap()->gc_threads_do(&cl);
-  if (StringDedup::is_enabled()) {
-    StringDedup::threads_do(&cl);
-  }
-  cl.do_thread(WatcherThread::watcher_thread());
-  cl.do_thread(AsyncLogWriter::instance());
+  non_java_threads_do(&cl);
 
   // SapMachine 2019-11-07: Vitals
```
reinrich: I think L3868 - L3883 can be also removed. The vitals sampler thread and the himem reporter thread are `NonJavaThread`s and handled by `non_java_threads_do(&cl);` in L3866.

RealCLanger: Interesting, good catch. Then it's probably also wrong in head and the other higher backports, e.g.: https://github.com/SAP/SapMachine/blob/32cfd1865848364f004da243636a3951c188754a/src/hotspot/share/runtime/threads.cpp#L1389 correct?

SapMachine:

src/hotspot/share/runtime/thread.cpp
reinrich: Yes indeed. Likewise good catch.

src/hotspot/share/runtime/thread.cpp
```diff
@@ -3589,7 +3589,6 @@ void Threads::add(JavaThread* p, bool force_daemon) {
   p->set_on_thread_list();
 
   _number_of_threads++;
-
```
reinrich, line 3592: Other nit: empty line L3618 before closing brace doesn't look nice either. I'll leave it to you if you remove it.

reinrich (approved):

src/hotspot/share/runtime/thread.cpp
RealCLanger: I removed this line only because it's a diff to upstream. 😉

reinrich: Yeah I noticed it.

PR #1765, opened 2024-07-25
Disable use of backtrace() on PPC platforms by default

schmelter-sap:
It can crash in the backtrace() method on newer versions of glibc when tls code is on the stack.

Additionally a missing disenroll() call was added to the delayed enabling task.

fixes #1550


SapMachine:

RealCLanger (approved):

reinrich (approved):
Looks good.
Cheers, Richard.

PR #1794, opened 2024-08-19
Don't try to resolve CodeBlobs in malloc trace dumps

### LINKED ISSUE #1789: Don't try to resolve CodeBlobs in malloc trace dumps
Since nmethods might get removed or deleted, just using the PC of an old stacktrace could lead to getting the wrong CodeBlob. This is probably more confusing than not having the information at all.

schmelter-sap:
This removes the overoptimistic feature to get the code blob of a PC on the stack when printing stack traces. This could lead to misleading methods being printed, when the blob was moved or deleted in the meantime.

fixes #1789

### DIFF
--- a/src/hotspot/os/posix/malloctrace/mallocTracePosix.cpp
+++ b/src/hotspot/os/posix/malloctrace/mallocTracePosix.cpp
@@ -30,7 +30,6 @@
 #include "mallochooks.h"
 #include "malloctrace/mallocTracePosix.hpp"
 
-#include "code/codeBlob.hpp"
 #include "code/codeCache.hpp"
 #include "runtime/arguments.hpp"
 #include "runtime/atomic.hpp"
@@ -1387,12 +1386,8 @@ void MallocStatisticImpl::record_allocation(void* ptr, uint64_t hash, int nr_of_
         if (os::print_function_and_library_name(&ss, frame, tmp, sizeof(tmp), true, true, false)) {
           ss.cr();
         } else {
-          CodeBlob* blob = CodeCache::find_blob((void*) frame);
-
-          if (blob != nullptr) {
-            ss.print_raw(" ");
-            blob->print_value_on(&ss);
-            ss.cr();
+          if ((frame >= CodeCache::low_bound()) && (frame < CodeCache::high_bound())) {
+            ss.print_raw_cr(" <code cache>");
           } else {
             ss.print_raw_cr(" <unknown code>");
           }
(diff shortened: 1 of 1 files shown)


SapMachine:

SapMachine:

reinrich (commented):
Except for PPC `-XX:+MallocTraceUseBacktrace` is the default. `backtrace()` cannot capture the java part of the stack, can it?

With `-XX:-MallocTraceUseBacktrace` the [frames with a code blobe are excluded](https://github.com/reinrich/SapMachine/blob/00cf7d242f74519b3f33ae050a9bcabbad8fb509/src/hotspot/os/posix/malloctrace/mallocTracePosix.cpp#L861) anyway. Am I correct, and if so: is this a bug or a feature?

schmelter-sap:
The non-backtrace code also stops at the first frame with a code blob associated (due to the fr.cb() != nullptr check in MallocStatisticImpl::capture_stack()).

reinrich:
> The non-backtrace code also stops at the first frame with a code blob associated (due to the fr.cb() != nullptr check in MallocStatisticImpl::capture_stack()).

That's what I meant. I wasn't sure if this was the desired behavior.
So (at most) one frame with an associated code blob is added to the stack trace. Its callers are not added. And that's on purpose. The malloc trace is not meant for analyzing native memory problems originating from java. E.g. leaks where memory is allocated with `Unsafe::allocateMemory` cannot be analyzed, correct?

schmelter-sap:
That's correct. We would see that a memory leak was caused by `Unsafe::allocateMemory`, but would not see the Java code allocating the leaked memory. The same can be said for direct byte buffers.

src/hotspot/os/posix/malloctrace/mallocTracePosix.cpp
```diff
@@ -1843,11 +1838,8 @@ static void print_frame(outputStream* st, address frame) {
   if (os::print_function_and_library_name(st, frame, tmp, sizeof(tmp), true, true, false)) {
     st->cr();
   } else {
-    CodeBlob* blob = CodeCache::find_blob((void*) frame);
-
-    if (blob != nullptr) {
-      st->print_raw(" ");
-      blob->print_value_on(st);
+    if ((frame >= CodeCache::low_bound()) && (frame < CodeCache::high_bound())) {
+      st->print_raw_cr(" <code cache>");
```
reinrich, line 1846: Please add a short comment why a lookup of the code blob for the pc wouldn't always yield the correct result.

reinrich (approved):
Looks good to me. A short comment would be nice though.
Thanks, Richard.

SapMachine:

reinrich (approved):
Thanks for adding the comment.

RealCLanger (approved):

PR #1811, opened 2024-09-05
Make test for unique stacks more robust

### LINKED ISSUE #1810: Make test for unique stacks more robust
This test fails with an assertion in the profiled VM, caused by the buffer used to store the output requested by jcmd.

schmelter-sap:
This changes the way the stack traces are routed to the test the uniques stacks test, since it produces a big output. We now let the profiled VM print the stacks to stderr and not use the output of jcmd, since the latter is buffered in the profiled VM first and there are hard limits on the maximum size of that buffer,.

fixes #1810

### DIFF
--- a/test/hotspot/jtreg/runtime/malloctrace/MallocHooksTest.java
+++ b/test/hotspot/jtreg/runtime/malloctrace/MallocHooksTest.java
@@ -35,7 +35,9 @@ import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.InputStreamReader;
+import java.io.IOException;
 import java.io.StringReader;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


SapMachine:

test/hotspot/jtreg/runtime/malloctrace/MallocHooksTest.java
```diff
@@ -595,11 +597,25 @@ private static void testUniqueStacks() throws Exception {
                                   "-XX:MallocTraceStackDepth=14").start();
         checkIsAttachable(p);
         p.getInputStream().read();
-        OutputAnalyzer oa = callJcmd(p, "MallocTrace.dump", "-percentage=100");
+        // The output can be large, so dump it directly, since the output of
+        // jcmd is temporary hold in the JVM and there is a limit of 100 MB.
+        final String[] stacks = new String[1];
```
RealCLanger, line 602: Any reason why you are using a String[] here instead of a single String object?

RealCLanger (commented):
This change makes sense, I guess. Let's see whether all tests are green.

test/hotspot/jtreg/runtime/malloctrace/MallocHooksTest.java
schmelter-sap: You cannot modify a local variable in an anonymous inner class. So the array trick is the common way.

RealCLanger (approved):
OK. We should port this fix to upstream, too, I guess?

schmelter-sap:
> OK. We should port this fix to upstream, too, I guess?

If this really fixes the problem in the tests.

RealCLanger:
> > OK. We should port this fix to upstream, too, I guess?
> 
> If this really fixes the problem in the tests.

It should - it looks greenish now. And the error happened on all our ppc64le nodes before.

MBaesken (approved):

PR #1953, opened 2025-04-18
Merge to tag jdk-25+19

SapMachine:
please pull


TheRealMDoerr (approved):
Only timeouts in new test TestFloat16VectorOperations on PPC64. We may be able to implement some Float16VectorOperations in the future, but only for newer Power processors.

MBaesken (approved):

PR #1962, opened 2025-05-05
Don't use libc allocations directly in malloc trace

### LINKED ISSUE #1961: Don't use libc allocations directly in malloc trace
Since the OpenJDK wants to remove the usage of 'naked' memory allocation, its use in the malloc trace implementation should be adjusted. While the memory allocation functions themselves are not called, the addresses of the methods are used to insert them into the stack trace instead of the hooks which were called in reality. This should not be done anymore (and it wasn't possible on MacOSX anyway).

schmelter-sap:
Removes the use of 'naked' allocation functions in the malloc trace. This is in preparation to openjdk removing the access to these.

fixes #1961

### DIFF
--- a/src/hotspot/os/posix/malloctrace/mallocTracePosix.cpp
+++ b/src/hotspot/os/posix/malloctrace/mallocTracePosix.cpp
@@ -40,10 +40,6 @@
 #include <pthread.h>
 #include <stdlib.h>
 
-#if !defined(__APPLE__)
-#include <malloc.h>
-#endif
-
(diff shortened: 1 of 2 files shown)


SapMachine:

SapMachine:

MBaesken:
AIX and Linux ppc64le check failures seem to be unrelated , we have them in other builds too

MBaesken (approved):

RealCLanger (approved):

PR #2086, opened 2025-10-24
Merge to tag jdk-26+21

SapMachine:
please pull


TheRealMDoerr (approved):
I've resolved such that we use SapMachine 25 GA builds. We could switch to 25.0.1 in a separate PR if desired.

RealCLanger:
> I've resolved such that we use SapMachine 25 GA builds. We could switch to 25.0.1 in a separate PR if desired.

I'd say that we should use SapMachine 25.0.1 as it is available. For MacOS x86 you could use the build from https://github.com/SAP/SapMachine/releases/tag/sapmachine-25%2B36.

TheRealMDoerr:
> I'd say that we should use SapMachine 25.0.1 as it is available.

I've changed to 25.0.1, but some jobs were aborted.

> For MacOS x86 you could use the build from https://github.com/SAP/SapMachine/releases/tag/sapmachine-25%2B36.

I prefer keeping MacOS x86 as it was tested by other people. We have no usage for it, so it would be an unnecessary diff.

RealCLanger:
OK, let's wait for the missing PR validation jobs.

RealCLanger (approved):
OK, two PR validation jobs (windows and linux alpine) failing due to new environment. Let's work on that with next PR.

PR #2110, opened 2025-11-24
SapMachine #2109: Linux: Configuration of glibc malloc arenas should be possible

### LINKED ISSUE #2109: Linux: Configuration of glibc malloc arenas should be possible
We often have issues with glibc malloc arenas (see https://bugs.openjdk.org/browse/JDK-8193521). We should ideally switch them off by default (limit to 1), but make that configurable in case anybody needs it differently. Several of our users have reported issues with the default which is 8 * CPU count. No user has reported issues related to malloc/free performance when using only 1 malloc arena.

Alternatively, it is possible to set the "MALLOC_ARENA_MAX" environment variable, but that is much harder for our use cases than having a switch with the default such that it prevents high memory utilization due to malloc arenas.

TheRealMDoerr:
Make glibc malloc arenas configurable, but use only 1 by default. See Issue below.

fixes #2109

### DIFF
--- a/src/hotspot/os/linux/globals_linux.hpp
+++ b/src/hotspot/os/linux/globals_linux.hpp
@@ -88,6 +88,10 @@
          "omitted.\n"                                                  \
          "Example: \"-XX:HiMemReportExec=GC.class_histogram -all;GC.heap_dump\"") \
                                                                         \
+  /* SapMachine 2025-11-24: Configurable limit of malloc arenas */      \
+  product(int, GlibcMallocArenas, 1,                                    \
+          "Limit glibc malloc arenas, 0 means use OS default")          \
+                                                                        \
   product(bool, UseCpuAllocPath, false, DIAGNOSTIC,                     \
           "Use CPU_ALLOC code path in os::active_processor_count ")     \
 																	                                      \
--- a/src/hotspot/os/linux/globals_linux.hpp
+++ b/src/hotspot/os/linux/globals_linux.hpp
@@ -90,11 +90,12 @@
                                                                         \
   /* SapMachine 2025-11-24: Configurable limit of malloc arenas */      \
   product(int, GlibcMallocArenas, 1,                                    \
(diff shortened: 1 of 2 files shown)


SapMachine:
Hello @TheRealMDoerr, I'm sorry, this pull request doesn't meet the expectations. The pull request description has an invalid format. Please have a look at https://github.com/SAP/SapMachine/wiki/Formal-Requirements-of-Pull-Requests .

src/hotspot/os/linux/os_linux.cpp
ansteiner, line 1: From my current investigation the malloc arena is 64MB on Linux x86_64.

src/hotspot/os/linux/globals_linux.hpp
```diff
@@ -88,6 +88,10 @@
          "omitted.\n"                                                  \
          "Example: \"-XX:HiMemReportExec=GC.class_histogram -all;GC.heap_dump\"") \
                                                                         \
+  /* SapMachine 2025-11-24: Configurable limit of malloc arenas */      \
+  product(int, GlibcMallocArenas, 1,                                    \
+          "Limit glibc malloc arenas, 0 means use OS default")          \
```
MBaesken: maybe add - 'our default is only one arena for minimum memory use.'

src/hotspot/os/linux/os_linux.cpp
MBaesken: Seems this can vary, based on distro, glibc version, CPUs etc. etc. .

TheRealMDoerr:
Thanks for reviewing! I've improved the description and the comment.

RealCLanger:
retest this please

SapMachine:
Hello @TheRealMDoerr, this pull request fulfills all formal requirements.

MBaesken (approved):

RealCLanger (approved):
LGTM. Please squash when merging.

PR #2111, opened 2025-11-25
SapMachine (25) #2109: Linux: Configuration of glibc malloc arenas should be possible

### LINKED ISSUE #2109: Linux: Configuration of glibc malloc arenas should be possible

TheRealMDoerr:
Clean backport of https://github.com/SAP/SapMachine/pull/2110 (cherry-picked).

fixes #2109

### DIFF
--- a/src/hotspot/os/linux/globals_linux.hpp
+++ b/src/hotspot/os/linux/globals_linux.hpp
@@ -91,9 +91,14 @@
          "omitted.\n"                                                  \
          "Example: \"-XX:HiMemReportExec=GC.class_histogram -all;GC.heap_dump\"") \
                                                                         \
+  /* SapMachine 2025-11-24: Configurable limit of malloc arenas */      \
+  product(int, GlibcMallocArenas, 1,                                    \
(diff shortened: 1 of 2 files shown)


SapMachine:
Hello @TheRealMDoerr, this pull request fulfills all formal requirements.

SapMachine:
Hello @TheRealMDoerr, this pull request fulfills all formal requirements.

RealCLanger (approved):

ArnoZeller (approved):
LGTM. 
The different test failures are either  known or environment related.

TheRealMDoerr:
Thanks for the reviews!

PR #2148, opened 2025-12-25
Merge to tag jdk-21.0.10+6

SapMachine:
please pull


TheRealMDoerr (approved):
The email issue can't be solved because the change is coming from OpenJDK and was hence contributed with OpenJDK email address.

RealCLanger (approved):

PR #2149, opened 2025-12-25
Merge to tag jdk-17.0.18+7

SapMachine:
please pull


TheRealMDoerr (approved):
Assertion on linux x64_64 is [JDK-8370200](https://bugs.openjdk.org/browse/JDK-8370200) which is unrelated to this merge PR.
The email issue can't be solved because the change is coming from OpenJDK and was hence contributed with OpenJDK email address.

RealCLanger (approved):

PR #2162, opened 2026-01-28
SapMachine (27) #2161:  CheckFiles jtreg tests fails in SapMachine 27

### LINKED ISSUE #2161: CheckFiles jtreg tests fails in SapMachine 27
It complains about a few jars in lib
```
  visiting file:/priv/jvmtests/output_sapmachine27_release_optU_linuxppc64le/testee-vm/lib/agent-boot.jar
  --> ERROR this file is not allowed:agent-boot.jar
  visiting file:/priv/jvmtests/output_sapmachine27_release_optU_linuxppc64le/testee-vm/lib/agent.jar
  --> ERROR this file is not allowed:agent.jar
```

and there is also additional stuff in legal/async .
This is expected in SapMachine , because we add async-profiler on some platforms.

MBaesken:
Relax the check of the lib folder, because we might include async-profiler jars

fixes #2161

### DIFF
--- a/test/jdk/build/CheckFiles.java
+++ b/test/jdk/build/CheckFiles.java
@@ -84,7 +84,9 @@ public class CheckFiles {
         allowedEndingsLibDir.add("fontconfig.bfc");
         allowedEndingsLibDir.add("fontconfig.properties.src");
         allowedEndingsLibDir.add("ct.sym");
-        allowedEndingsLibDir.add("jrt-fs.jar");
+        // SapMachine 2026-01-28 allow all jars because of our added async-profiler
+        //allowedEndingsLibDir.add("jrt-fs.jar");
+        allowedEndingsLibDir.add(".jar");
         allowedEndingsLibDir.add("jvm.cfg");
         allowedEndingsLibDir.add("modules");
         allowedEndingsLibDir.add("psfontj2d.properties");


SapMachine:
Hello @MBaesken, this pull request fulfills all formal requirements.

test/jdk/build/CheckFiles.java
```diff
@@ -84,7 +84,9 @@ public static void main(String[] args) throws Exception {
         allowedEndingsLibDir.add("fontconfig.bfc");
         allowedEndingsLibDir.add("fontconfig.properties.src");
         allowedEndingsLibDir.add("ct.sym");
-        allowedEndingsLibDir.add("jrt-fs.jar");
+        // SapMachine 2026-01-28 allow all jars because of our added async-profiler
+        //allowedEndingsLibDir.add("jrt-fs.jar");
```
RealCLanger, line 88: Hm, shouldn't we be more specific here and list all expected jars by name?

MBaesken: This test is only a 'rough' check and not comparable to the  SAPJVM file lists checks.  SO  just checking '.jar' follows the idea of this specific test.

MBaesken: (and if we want to check exact files, we have to consider that they are 1.  platform specific and 2.  not even  there  imho  just after a local  Sapmachine developers  build )

RealCLanger (approved):

ArnoZeller (approved):
Looks fine for me.

PR #2166, opened 2026-02-02
SapMachine 21 #2165: G1 should initiate the next concurrent cycle even before finishing a mixed phase if there are excessive hum. allocations

### LINKED ISSUE #2165: G1: Sudden Full GC after longer Phase without any GC Activity
We have received reports for SapMachine 21 about G1 full gcs that are rather unexpected from the user's perspective since (a) the full gcs are initiated after a longer phases without any gc activity and (b) the full gcs reveal that the heap has filled up with objects of which most are unreachable.

This [excerpt of gc logs](https://github.com/user-attachments/files/25025123/SF_fullGC_behavior.txt) shows this pattern (visualized in a [chart](https://github.com/user-attachments/files/25066188/SF_fullGC_behavior.pdf)).
In this example there's a phase of 30s without gc activity and then a full gc is triggered.
According to the data the heap fills up with short lived humongous allocations.
Note that eden has been expanded by 3x to 9 GB in GC 12054 (region size 32 MB) just before the phase without gc activity. After the concurrent marking started with 12052 G1 entered a mixed phase where it performs mixed gcs only. Due to the eden expansion it takes long until the next mixed evacuation. Meanwhile free regions were consumed by humongous allocations and evacuation fails.

The behaviour resembles

[8355972:](https://bugs.openjdk.org/browse/JDK-8355972) G1: Exclusive humongous allocations trigger unnecessary Full GC

As explained above the issue can also occur if eden is very large. 

G1 should schedule a concurrent cycle to reclaim unreachable humongous objects. This would likely prevent the full gc.

reinrich:
**UPDATE**: The new option was renamed to `G1MixedGCHumongousAllocThreshold`.

This pr introduces a new experimental option `G1HumAllocPercentUntilConcurrent` that controls the G1 enhancement described below. With its default value 0 the enhancement is disabled.

A value >0 is interpreted as threshold for humongous allocations given as percentage of the maximum heap size. The threshold limits humongous allocations between mixed evacuation pauses. If reached the current mixed phase is aborted in favor of a new concurrent marking cycle.

Without this enhancement G1 would not initiate concurrent marking in a mixed phase. The problem with this behavior is that the heap can fill up with humongous allocations while G1 waits for the next mixed evacuation pause that will be scheduled only when eden is full. If the heap is full of humongous objects before then a full gc is necessary.

_Testing_

The reproducer from

> [8355972:](https://bugs.openjdk.org/browse/JDK-8355972) G1: Exclusive humongous allocations trigger unnecessary Full GC

doesn't show full gcs if the enhancement is enabled with `-XX:+UnlockExperimentalVMOptions -XX:G1HumAllocPercentUntilConcurrent=30`

Full command:

```
java -Xmx1g -Xlog:gc* -XX:+UnlockExperimentalVMOptions -XX:G1HumAllocPercentUntilConcurrent=30 HumongousGc.java
```

The change includes a regression test similar to the reproducer.

A patch with G1HumAllocPercentUntilConcurrent=10 is part of our nightly testing since January 30.

Also tested with https://github.com/reinrich/TestGC

```
./images/jdk/bin/java
-Xmx8g
-XX:+UseG1GC
-XX:+UnlockExperimentalVMOptions -XX:G1MixedGCHumongousAllocThreshold=5
test.gc.MAIN MEDIUM HUM_MEDIUM HUM_ONLY_MEDIUM
```

fixes #2165

### DIFF
--- a/src/hotspot/share/gc/g1/g1Policy.cpp
+++ b/src/hotspot/share/gc/g1/g1Policy.cpp
@@ -73,6 +73,8 @@ G1Policy::G1Policy(STWGCTimer* gc_timer) :
   _rs_length(0),
   _pending_cards_at_gc_start(0),
   _concurrent_start_to_mixed(),
+  // SapMachine 2026-01-20: force G1 marking in mixed phase in case of excessive hum. allocations
+  _max_hum_bytes_in_mixed_phase(0),
   _collection_set(nullptr),
   _g1h(nullptr),
   _phase_times_timer(gc_timer),
@@ -106,6 +108,9 @@ void G1Policy::init(G1CollectedHeap* g1h, G1CollectionSet* collection_set) {
   // We immediately start allocating regions placing them in the collection set.
   // Initialize the collection set info.
   _collection_set->start_incremental_building();
+
+  // SapMachine 2026-01-20: force G1 marking in mixed phase in case of excessive hum. allocations
+  _max_hum_bytes_in_mixed_phase = _g1h->reserved().byte_size() * G1HumAllocPercentUntilConcurrent / 100;
 }
 
 void G1Policy::record_young_gc_pause_start() {
@@ -700,8 +705,30 @@ bool G1Policy::about_to_start_mixed_phase() const {
   return _g1h->concurrent_mark()->cm_thread()->in_progress() || collector_state()->in_young_gc_before_mixed();
 }
 
+// SapMachine 2026-01-20: force G1 marking in mixed phase in case of excessive hum. allocations
+bool G1Policy::force_concurrent_ending_mixed_phase(bool hum_alloc) {
+  // Only humongous allocations are subject to the policy. It applies if a maximum > 0 is given.
+  bool policy_applies =
+      (_max_hum_bytes_in_mixed_phase > 0) && hum_alloc && !_g1h->concurrent_mark()->cm_thread()->in_progress();
+
+  // The collector might not honor marking requests if expecting mixed collections.
+  // Force in case of excessive humongous allocations.
+  bool force_cm = policy_applies &&
+      (collector_state()->in_young_gc_before_mixed() || !collector_state()->in_young_only_phase()) &&
+      (old_gen_alloc_tracker()->allocated_humongous_bytes_since_last_gc() > _max_hum_bytes_in_mixed_phase);
+
+  if (force_cm) {
+    log_info(gc, ergo, ihop)("Excessive humongous allocations in mixed phase: " SIZE_FORMAT "B since last gc threshold:" SIZE_FORMAT "B",
+                             old_gen_alloc_tracker()->allocated_humongous_bytes_since_last_gc(), _max_hum_bytes_in_mixed_phase);
+  }
+
+  return force_cm;
+}
+
 bool G1Policy::need_to_start_conc_mark(const char* source, size_t alloc_word_size) {
-  if (about_to_start_mixed_phase()) {
+  // SapMachine 2026-01-20: force G1 marking in mixed phase in case of excessive hum. allocations
+  bool should_force_cm = force_concurrent_ending_mixed_phase(_g1h->is_humongous(alloc_word_size));
+  if (about_to_start_mixed_phase() && !should_force_cm) {
     return false;
   }
 
@@ -714,8 +741,11 @@ bool G1Policy::need_to_start_conc_mark(const char* source, size_t alloc_word_siz
   bool result = false;
   if (marking_request_bytes > marking_initiating_used_threshold) {
     result = collector_state()->in_young_only_phase() && !collector_state()->in_young_gc_before_mixed();
-    log_debug(gc, ergo, ihop)("%s occupancy: " SIZE_FORMAT "B allocation request: " SIZE_FORMAT "B threshold: " SIZE_FORMAT "B (%1.2f) source: %s",
+    // SapMachine 2026-01-20: force G1 marking in mixed phase in case of excessive hum. allocations
+    result = result || should_force_cm;
+    log_debug(gc, ergo, ihop)("%s%s occupancy: " SIZE_FORMAT "B allocation request: " SIZE_FORMAT "B threshold: " SIZE_FORMAT "B (%1.2f) source: %s",
                               result ? "Request concurrent cycle initiation (occupancy higher than threshold)" : "Do not request concurrent cycle initiation (still doing mixed collections)",
+                              should_force_cm ? "(ending mixed phase)" : "",
                               cur_used_bytes, alloc_byte_size, marking_initiating_used_threshold, (double) marking_initiating_used_threshold / _g1h->capacity() * 100, source);
   }
   return result;
@@ -1213,7 +1243,9 @@ void G1Policy::decide_on_concurrent_start_pause() {
     } else if (_g1h->is_user_requested_concurrent_full_gc(cause) ||
                (cause == GCCause::_codecache_GC_threshold) ||
                (cause == GCCause::_codecache_GC_aggressive) ||
-               (cause == GCCause::_wb_breakpoint)) {
+               (cause == GCCause::_wb_breakpoint) ||
+               // SapMachine 2026-01-20: force G1 marking in mixed phase in case of excessive hum. allocations
+               (force_concurrent_ending_mixed_phase(cause == GCCause::_g1_humongous_allocation))) {
       // Initiate a concurrent start.  A concurrent start must be a young only
       // GC, so the collector state must be updated to reflect this.
       collector_state()->set_in_young_only_phase(true);
@@ -1226,8 +1258,8 @@ void G1Policy::decide_on_concurrent_start_pause() {
       abandon_collection_set_candidates();
       abort_time_to_mixed_tracking();
       initiate_conc_mark();
-      log_debug(gc, ergo)("Initiate concurrent cycle (%s requested concurrent cycle)",
-                          (cause == GCCause::_wb_breakpoint) ? "run_to breakpoint" : "user");
+      // SapMachine 2026-01-20: force G1 marking in mixed phase in case of excessive hum. allocations
+      log_debug(gc, ergo)("Initiate concurrent cycle (%s)", GCCause::to_string(cause));
     } else {
       // The concurrent marking thread is still finishing up the
       // previous cycle. If we start one right now the two cycles
--- a/src/hotspot/share/gc/g1/g1Policy.cpp
+++ b/src/hotspot/share/gc/g1/g1Policy.cpp
@@ -1258,7 +1258,12 @@ void G1Policy::decide_on_concurrent_start_pause() {
       abort_time_to_mixed_tracking();
       initiate_conc_mark();
       // SapMachine 2026-01-20: force G1 marking in mixed phase in case of excessive hum. allocations
-      log_debug(gc, ergo)("Initiate concurrent cycle (%s)", GCCause::to_string(cause));
+      if (cause == GCCause::_g1_humongous_allocation) {
+        log_debug(gc, ergo)("Initiate concurrent cycle (%s)", GCCause::to_string(cause));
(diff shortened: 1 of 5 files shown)


SapMachine:
Hello @reinrich, this pull request fulfills all formal requirements.

test/hotspot/jtreg/gc/g1/TestHumongousAllocMixedPhase.java
```diff
@@ -0,0 +1,98 @@
...
+public class TestHumongousAllocMixedPhase {
+    // Heap sizes < 224 MB are increased to 224 MB if vm_page_size == 64K to
+    // fulfill alignment constraints.
+    private static final int heapSize                       = 224; // MB
+    private static final int heapRegionSize                 = 1;   // MB
+
+    public static void main(String[] args) throws Exception {
+        boolean success = false;
+        int allocDelayMsMax = 500;
+        for(int allocDelayMs = 10; !success; allocDelayMs *= 4) {
```
RealCLanger:
```suggestion
        for (int allocDelayMs = 10; !success; allocDelayMs *= 4) {
```

src/hotspot/share/gc/g1/g1_globals.hpp
```diff
@@ -323,6 +323,15 @@
           "related prediction sample. That sample must involve the same or "\
           "more than that number of cards to be used.")                     \
                                                                             \
+  /* SapMachine 2026-01-20 */                                               \
+  product(uintx, G1HumAllocPercentUntilConcurrent, 0, EXPERIMENTAL,         \
+          "Threashold for humongous allocations since last gc to initiate a"\
```
RealCLanger:
```suggestion
          "Threshold for humongous allocations since last gc to initiate a "\
```

src/hotspot/share/gc/g1/g1_globals.hpp
```diff
@@ -323,6 +323,15 @@
...
+          "new concurrent cycle even if the collector is not yet done with" \
```
RealCLanger:
```suggestion
          "new concurrent cycle even if the collector is not yet done with "\
```

src/hotspot/share/gc/g1/g1_globals.hpp
```diff
@@ -323,6 +323,15 @@
...
+          "mixed collections. This can help in longer phases where no"      \
```
RealCLanger:
```suggestion
          "mixed collections. This can help in longer phases where no "     \
```

src/hotspot/share/gc/g1/g1_globals.hpp
```diff
@@ -323,6 +323,15 @@
...
+          "evacuation pauses are scheduled because most allocations are"    \
```
RealCLanger:
```suggestion
          "evacuation pauses are scheduled because most allocations are "   \
```

src/hotspot/share/gc/g1/g1_globals.hpp
```diff
@@ -323,6 +323,15 @@
...
+          "humongous.  The value is a percentage of the maximum heap size.")\
```
RealCLanger:
```suggestion
          "humongous. The value is a percentage of the maximum heap size.") \
```

RealCLanger (changes requested):
Only a formal review, not a technical one. For that, @schmelter-sap or @TheRealMDoerr, please have a look. Thx.

reinrich:
Thanks Christoph. I've pushed the requested changes.

SapMachine:
Hello @reinrich, this pull request fulfills all formal requirements.

src/hotspot/share/gc/g1/g1_globals.hpp
```diff
@@ -323,6 +323,15 @@
           "related prediction sample. That sample must involve the same or "\
           "more than that number of cards to be used.")                     \
                                                                             \
+  /* SapMachine 2026-01-20 */                                               \
+  product(uintx, G1HumAllocPercentUntilConcurrent, 0, EXPERIMENTAL,         \
```
TheRealMDoerr: I think this name is only understandable for G1 nerds ;-)
Would be great if the name could indicate that we start a new concurrent full GC after the threshold has been reached. Maybe `G1HumPercentConcFullGCThreshold`?

test/hotspot/jtreg/gc/g1/TestHumongousAllocMixedPhase.java
```diff
@@ -0,0 +1,98 @@
...
+ */
+
+import jdk.test.lib.process.OutputAnalyzer;
+import jdk.test.lib.process.ProcessTools;
+
+public class TestHumongousAllocMixedPhase {
+    // Heap sizes < 224 MB are increased to 224 MB if vm_page_size == 64K to
+    // fulfill alignment constraints.
+    private static final int heapSize                       = 224; // MB
+    private static final int heapRegionSize                 = 1;   // MB
```
TheRealMDoerr: I'd use fewer whitespaces.

src/hotspot/share/gc/g1/g1Policy.cpp
```diff
@@ -700,8 +705,30 @@ bool G1Policy::about_to_start_mixed_phase() const {
   return _g1h->concurrent_mark()->cm_thread()->in_progress() || collector_state()->in_young_gc_before_mixed();
 }
 
+// SapMachine 2026-01-20: force G1 marking in mixed phase in case of excessive hum. allocations
+bool G1Policy::force_concurrent_ending_mixed_phase(bool hum_alloc) {
```
schmelter-sap: The method name makes it look like it does ending the phase instead of just deciding if it should be stopped.

src/hotspot/share/gc/g1/g1Policy.cpp
```diff
@@ -1213,7 +1243,9 @@ void G1Policy::decide_on_concurrent_start_pause() {
     } else if (_g1h->is_user_requested_concurrent_full_gc(cause) ||
                (cause == GCCause::_codecache_GC_threshold) ||
                (cause == GCCause::_codecache_GC_aggressive) ||
-               (cause == GCCause::_wb_breakpoint)) {
+               (cause == GCCause::_wb_breakpoint) ||
+               // SapMachine 2026-01-20: force G1 marking in mixed phase in case of excessive hum. allocations
+               (force_concurrent_ending_mixed_phase(cause == GCCause::_g1_humongous_allocation))) {
```
schmelter-sap: It seems the hum_alloc boolean argument in the call to force_concurrent_ending_mixed_phase is only needed for this code.
You could use ?: here and remove the argument form the call, since hum_alloc being false just means the method returns false.

src/hotspot/share/gc/g1/g1Policy.cpp
```diff
@@ -1226,8 +1258,8 @@ void G1Policy::decide_on_concurrent_start_pause() {
       abandon_collection_set_candidates();
       abort_time_to_mixed_tracking();
       initiate_conc_mark();
-      log_debug(gc, ergo)("Initiate concurrent cycle (%s requested concurrent cycle)",
-                          (cause == GCCause::_wb_breakpoint) ? "run_to breakpoint" : "user");
+      // SapMachine 2026-01-20: force G1 marking in mixed phase in case of excessive hum. allocations
+      log_debug(gc, ergo)("Initiate concurrent cycle (%s)", GCCause::to_string(cause));
```
schmelter-sap: Wouldn't that change the log output for all kinds of GC causes?

schmelter-sap (approved):
Some more cosmetic remarks. Looks good otherwise.

src/hotspot/share/gc/g1/g1_globals.hpp
dbriemann: I would wager that anyone who does not know the Parameter has to read the doc anyways. So stuffing in every related keyword doesn't really help.

Thus I would suggest something more readable like `G1HumongousAllocThreshold`.

reinrich:
Thanks for the reviews! Please have a look at the commits where I addressed them.

src/hotspot/share/gc/g1/g1_globals.hpp
reinrich: What about `G1MixedGCHumongousAllocThreshold`? That name is similar to other options above.

SapMachine:
Hello @reinrich, this pull request fulfills all formal requirements.

src/hotspot/share/gc/g1/g1Policy.cpp
reinrich: In `need_to_start_conc_mark()` I wanted to make the call only if called by `attempt_allocation_humongous()` in order to change the behavior as little as possible. I've changed the callers to make the call only if a hum. alloc. is going on. Is that ok?

src/hotspot/share/gc/g1/g1Policy.cpp
reinrich: But it's incorrect anyway. The causes related to the code cache are logged as "user requested". I can revert my change and print s.th. different for cause == GCCause::_g1_humongous_allocation. Shall I?

src/hotspot/share/gc/g1/g1_globals.hpp
TheRealMDoerr: The Mixed GC is cancelled and a (concurrent) Full GC started, right?
`G1HumongousAllocFullGCThreshold` would be more comprehensive for me. Can this only happen if a Mixed GC is running or why would you want to have MixedGC in the name?

src/hotspot/share/gc/g1/g1Policy.cpp
TheRealMDoerr: Can this disturb anyone who is parsing the logs? Otherwise, the change looks fine to me.

src/hotspot/share/gc/g1/g1_globals.hpp
reinrich: The mixed gc phase (sequence of mixed evacuation pauses) is ended prematurely and a new cycle is initiated. Please have a look at [`G1MixedGCCountTarget`](https://github.com/openjdk/jdk/blob/37ae15a4896c700e0a47a43de3330e8879d147c2/src/hotspot/share/gc/g1/g1_globals.hpp#L290-L292) which is related. It's that phase of mixed gcs where G1 wants to make use of the liveness info computed by the marking to evacuate old regions with a lot of unreachable objects. G1 won't start new marking before it is done with the mixed gcs. I thought that `G1MixedGCHumongousAllocThreshold` matches the existing option and is also descriptive.

src/hotspot/share/gc/g1/g1Policy.cpp
reinrich: Could be. I've reverted my change but kept the new output for the new case `_g1_humongous_allocation`.

SapMachine:
Hello @reinrich, this pull request fulfills all formal requirements.

src/hotspot/share/gc/g1/g1_globals.hpp
TheRealMDoerr: Not the name I'd prefer from a Java developer's perspective, but ok. It's just a name of an experimental option and users can read the documentation which is good IMHO.

TheRealMDoerr (approved):
I think this is a helpful feature. The implementation LGTM. I've checked that the test fails without the fix and passes with it.

reinrich:
Thanks Martin. As mentioned in our daily I've done some further testing with another synthetic gc test that's hopefully a little bit more realistic than the reproducer from [8355972](https://bugs.openjdk.org/browse/JDK-8355972). The test fills the heap with short- and medium-lived and also with permanent objects and then keeps allocating - also humogous objects. I've adapted the test to pause allocations about every 10s except for humongous objects. This triggers stw full gcs. With the enhancement these can be avoided mostly.

RealCLanger (approved):

RealCLanger:
Please squash when merging. :)

PR #2171, opened 2026-02-06
Merge to tag jdk-27+8

SapMachine:
please pull


RealCLanger (approved):
Looks mergeable although alpine tests did not run (probably due to resource constraints) and there is an issue in JDK tier2 with test java/nio/channels/Selector/ConnectionRefusedMessage.java. But we can see whether it goes away with the next build.

TheRealMDoerr (approved):

PR #2219, opened 2026-04-15
Add method to check if a path in zipfs is a symbolic link

### LINKED ISSUE #2218: Add method to check if a path in zipfs is a symbolic link
Thus us requested by the XSA team.

schmelter-sap:
This adds a method to check if a given path from zipfs corresponds to a symbolic link.

fixes #2218


SapMachine:

RealCLanger:
restest this please

RealCLanger:
Do you really need this SharedSecrets stuff? I think for jdk.nio.zipfs it is fine to add public methods and call them since those are internal implementation classes.

You should only need to export the jdk.nio.zipfs package to jdk.sapext and then you can directly call any method you add.

src/jdk.sapext/share/classes/com/sap/jdk/ext/util/ZipfsUtils.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    /**
+     * Returns <code>true</code> if the given path is a {@link jdk.nio.zipfs.ZipPath} which
+     * represents a symbolic link.
+     *
+     * @param path The path in the zipfs.
+     * @return <code>true</code> if the path represents a symbolic link.
+     */
+    public static boolean isSymbolicLink(Object path) {
```
RealCLanger, line 43: The parameter should rather be java.nio.file.Path instead of just Object, no?

src/jdk.sapext/share/classes/com/sap/jdk/ext/util/ZipfsUtils.java
```diff
@@ -0,0 +1,52 @@
...
+        JdkNioZipfsAccess access = SharedSecrets.getJdkNioZipfsAccess();
+
+        if (access == null) {
+            return false;
+         }
+
+         return access.isSymbolicLink(path);
+    }
+ }
```
RealCLanger, line 52: Line break is missing

src/jdk.sapext/share/classes/com/sap/jdk/ext/util/ZipfsUtils.java
schmelter-sap: Probably yes.

schmelter-sap:
Regarding the shared secrets. A public method in an implementation class for which the user gets access too is still visible. Additionally I don't want to add a module dependency between the ext and the zipfs module, when it is not really needed.

SapMachine:

RealCLanger:
> Regarding the shared secrets. A public method in an implementation class for which the user gets access too is still visible. Additionally I don't want to add a module dependency between the ext and the zipfs module, when it is not really needed.

Sure but having this public method visible is no problem. jdk.nio.zipfs.ZipFileSystem is no public API class so nobody would see it as API. Obviously somebody might go and try to reflectively look it up and use it. But that's possible with any internal class and developers should avoid that because of compatibility risk. Also, in later JDKs reflective access is disabled by default or should be soon - not exactly sure where Java is at at the moment... 😄 

If you want to avoid a dependency between jdk.sapext and jdk.zipfs the way to go would also using services.

src/java.base/share/classes/jdk/internal/access/JdkNioZipfsAccess.java
```diff
@@ -0,0 +1,38 @@
...
+
+import java.nio.file.Path;
+
+/**
+ * SharedSecrets interface used for the access from java.text.Bidi
+ */
+
+public interface JdkNioZipfsAccess {
+
+    // java.awt.font.TextAttribute constants
```
RealCLanger: The comment is a bit bogus, I guess...

src/jdk.sapext/share/classes/com/sap/jdk/ext/util/ZipfsUtils.java
```diff
@@ -0,0 +1,54 @@
...
+import jdk.internal.access.SharedSecrets;
+ 
+public class ZipfsUtils {
+
+    // Silence warning about implicit ctor.
+    private ZipfsUtils() {
+    }
+
+    /**
+     * Returns <code>true</code> if the given path is a {@link jdk.nio.zipfs.ZipPath} which
```
RealCLanger: I think linking to jdk.nio.zipfs.ZipPath does not work here in Javadoc since jdk.zipfs is not imported here. Maybe remove the link.

src/jdk.zipfs/share/classes/jdk/nio/zipfs/ZipPath.java
```diff
@@ -62,6 +62,25 @@ final class ZipPath implements Path {
...
+            }
+
+            ZipPath zipPath = (ZipPath) path;
+            byte[] resolvedPath = zipPath.getResolvedPath();
+
+            return zipPath.zfs.isSymlink(resolvedPath);
+        }
+    }
+
+    // SapMachine 2026-04-14: Support for symlink detectiun.
```
RealCLanger:
```suggestion
    // SapMachine 2026-04-14: Support for symlink detection.
```

RealCLanger (changes requested):
OK, a few last minor suggestions...

src/java.base/share/classes/jdk/internal/access/JdkNioZipfsAccess.java
schmelter-sap: You guessed correctly.

src/jdk.sapext/share/classes/com/sap/jdk/ext/util/ZipfsUtils.java
schmelter-sap: Replaced it with a <code> tag.

src/java.base/share/classes/jdk/internal/access/JdkNioZipfsAccess.java
```diff
@@ -0,0 +1,37 @@
...
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package jdk.internal.access;
+
+import java.nio.file.Path;
+
+/**
+ * SharedSecrets interface used for the access from java.text.Bidi
```
RealCLanger:
```suggestion
 * SharedSecrets interface used for the access from jdk.nio.zipfs
```

RealCLanger (changes requested):

src/java.base/share/classes/jdk/internal/access/JdkNioZipfsAccess.java
schmelter-sap: Done

parttimenerd (approved):

RealCLanger (approved):

PR #2251, opened 2026-05-19
SapMachine #2250: Add adjustments to heap dumps for the buildpack

### LINKED ISSUE #2250: Add adjustments to heap dumps for the buildpack

schmelter-sap:
This change adds features to the heap dump for the cf buildpack to be able to store heap dumps for customers in the object store. It includes the following features:

- Added -XX:+HeapDumpOverwrite flag. This allows a heap dump to overwrite an existing file. It is also need to write to a named pipe. The latter will be used by the buildpack to stream from the named pipe to the object store. You could already do this in the jcmd GC.heap_dump command.
- Added -XX:HeapDumpParallelism=\<nr-of-threads> flag. This allows to specify the number of threads to use when writing the heap dump. You could already do this in the jcmd GC.heap_dump command.
- When using only one thread to dump the heap, the code will use additional files apart from the heap dump file. This ensures that no extra disk space is used when using only one thread.
- Added -XX:+LimitPrimitiveArrayContentInHeapDump flag. If enabled, the heap dump will only write the real content for the first elements of large primitive arrays. The rest is filled with 0 (or false for boolean arrays). This makes the heap dump much more compressible, making upload to the object store faster. The number of entries to keep can be set via -XX:StringLikeContentSizeLimitInHeapDump=\<limit> for byte and char arrays (default is 120) or -XX:ArrayContentSizeLimitInHeapDump=\<limit> for the other primitive arrays (default is 50).

fixes #2250


MBaesken:
Couple of minor adjustments might be needed in globals.hpp 
"the content of primitive arrays in not completely"    should be "is not"
"this only really safes space"   should be "saves"
In StringLikeContentSizeLimitInHeapDump description  -   "primitive char and bytes arrays"   should be "byte"

src/hotspot/share/services/heapDumper.cpp
```diff
@@ -1439,6 +1478,11 @@ void DumperSupport::dump_prim_array(AbstractDumpWriter* writer, typeArrayOop arr
     default : ShouldNotReachHere();
   }
 
+  // SAPJVM 2026-05-06: Fill with zeros, if we don't dump the whole content of the array.
+  if (fill_with_zero > 0) {
+    writer->write_zero((u4) fill_with_zero * type_size);
```
MBaesken, line 1484: Should this be size_t like it is used in the declaration of write_zero ?

src/hotspot/share/runtime/globals.hpp
```diff
@@ -721,6 +716,38 @@ const int ObjectAlignmentInBytes = 8;
...
+          "is enabled.")                                                    \
+          range(0, 100000)                                                  \
+                                                                            \
+  /* SAPJVM 2026-05-06: Allow to skip content of large arrays in dumps.*/   \
+  product(int, ArrayContentSizeLimitInHeapDump, 50, MANAGEABLE,             \
+          "The number of entries in a primitive array other than char and " \
+          "byte arrays to not skip in a heap dump when "                    \
+          "LimitPrimArrayContentInHeapDump is enabled.")                    \
+          range(0, 100000)                                                  \
+                                                                            \
```
MBaesken, line 764: Minor typos etc. see separate comment.

test/hotspot/jtreg/serviceability/HeapDump/PartialArrayContentTest.java
```diff
@@ -0,0 +1,246 @@
...
+        fos = new FileOutputStream(firstSegment);
+        fos.close();
+        createDump(dumpFile, false,
+                "-Xmx500M",
+                "-XX:+HeapDumpOnOutOfMemoryError",
+                "-XX:HeapDumpPath=" + dumpFile,
+                "-XX:+HeapDumpOverwrite",
+                "-XX:HeapDumpParallelism=1");
+        verifyDump(dumpFile, false);
+        Asserts.assertTrue(firstSegment.exists(), "Segment file should not have been created (parallelism=1).");
```
MBaesken, line 124: If firstSegments is NOT there (exists fails), assertTrue fails; so should the message be better something like 'should exist but does not' ?

test/hotspot/jtreg/serviceability/HeapDump/PartialArrayContentTest.java
```diff
@@ -0,0 +1,246 @@
...
+                    }
+                }
+            }
+
+            Asserts.assertTrue(expectedTypes.isEmpty());
+        }
+
+        return largestArraySize;
+    }
+}
```
MBaesken: Missing newline at file end

test/hotspot/jtreg/serviceability/HeapDump/PartialArrayContentTest.java
schmelter-sap: This tests if the segment file was created during the heap dump. It would overwrite the existing file and is deleted when the heap dump is merged. So if the segment file was created, the fristSegment file doesn't exist anymore.

test/hotspot/jtreg/serviceability/HeapDump/PartialArrayContentTest.java
schmelter-sap: Fixed

src/hotspot/share/runtime/globals.hpp
schmelter-sap: Fixed.

src/hotspot/share/services/heapDumper.cpp
schmelter-sap: The length in bytes in this function always uses u4. So for consistency with other code I kept it that way.

src/hotspot/share/runtime/globals.hpp
```diff
@@ -721,6 +716,38 @@ const int ObjectAlignmentInBytes = 8;
...
+          "If enabled, the heap dump on out of memory error can "           \
+          "overwrite an already existing file.")                            \
+                                                                            \
+  /* SapMachine 2026-05-06: Sets the parallelism of the heap dump. */       \
+  product(uint, HeapDumpParallelism, 0, MANAGEABLE,                         \
+          "Sets the parallelism of the heap dump creation. 0 means to let  "\
+          "the VM decide.")                                                 \
+                                                                            \
+  /* SAPJVM 2026-05-06: Allow to skip content of large arrays in dumps.*/   \
+  product(bool, LimitPrimArrayContentInHeapDump, false, MANAGEABLE,         \
```
dbriemann: Maybe consider the naming: `Primitive` instead of `Prim`. The name is long anyways so we can use a precise name.

src/hotspot/share/runtime/globals.hpp
```diff
@@ -721,6 +716,38 @@ const int ObjectAlignmentInBytes = 8;
...
+  product(bool, HeapDumpOverwrite, false, MANAGEABLE,                       \
+          "If enabled, the heap dump on out of memory error can "           \
+          "overwrite an already existing file.")                            \
+                                                                            \
+  /* SapMachine 2026-05-06: Sets the parallelism of the heap dump. */       \
+  product(uint, HeapDumpParallelism, 0, MANAGEABLE,                         \
+          "Sets the parallelism of the heap dump creation. 0 means to let  "\
+          "the VM decide.")                                                 \
+                                                                            \
+  /* SAPJVM 2026-05-06: Allow to skip content of large arrays in dumps.*/   \
```
dbriemann: Should the SAPJVM comments be SapMachine?

src/hotspot/share/services/heapDumper.cpp
```diff
@@ -1357,6 +1379,8 @@ void DumperSupport::dump_prim_array(AbstractDumpWriter* writer, typeArrayOop arr
 
   int length = calculate_array_max_length(writer, array, header_size);
   int type_size = type2aelembytes(type);
+  // SapMachine 2026-05-06
+  int fill_with_zero = 0;
```
dbriemann: We could move that to the next hunk (after line 1399), so we do not need this lonely comment, right?

src/hotspot/share/services/heapDumper.cpp
dbriemann: I agree with @MBaesken . This can overflow and the `write_zero` function takes `size_t` anyways. So why do this cast to `u4` here?

test/hotspot/jtreg/serviceability/HeapDump/PartialArrayContentTest.java
```diff
@@ -0,0 +1,246 @@
...
+        }
+    }
+    public static void main(String[] args) {
+        allocArrays();
+        LingeredApp.main(args);
+    }
+}
+
+class ArrayAllocOOMApp extends ArrayAllocApp {
+    public static int arraySize = 54321;
```
dbriemann: This is not needed / not used isn't it? The value used by `allocArrays()` seems to be in `ArrayAllocApp`.

src/hotspot/share/services/heapDumper.cpp
```diff
@@ -1372,6 +1396,21 @@ void DumperSupport::dump_prim_array(AbstractDumpWriter* writer, typeArrayOop arr
     return;
   }
 
+  // SAPJVM 2026-05-06: If enabled, we don't dump the whole content of large arrays, but just the start.
```
MBaesken: Still some 'SAPJVM' instead of SapMachine (at various source locations).

src/hotspot/share/services/heapDumper.cpp
schmelter-sap: It cannot overflow. The array length used is calculated via calculate_array_max_length(), which trims the length so length * type_size is always < 4GB. This is needed, since the packet length is a 32 bit unsigned int in the hprof format.

src/hotspot/share/runtime/globals.hpp
schmelter-sap: Done

src/hotspot/share/runtime/globals.hpp
schmelter-sap: Indeed.

src/hotspot/share/services/heapDumper.cpp
schmelter-sap: Fixed

src/hotspot/share/services/heapDumper.cpp
schmelter-sap: Done

test/hotspot/jtreg/serviceability/HeapDump/PartialArrayContentTest.java
schmelter-sap: Done

test/hotspot/jtreg/serviceability/HeapDump/PartialArrayContentTest.java
```diff
@@ -0,0 +1,273 @@
...
+import jdk.test.lib.Asserts;
+import jdk.test.lib.JDKToolLauncher;
+import jdk.test.lib.apps.LingeredApp;
+import jdk.test.lib.process.ProcessTools;
+import jdk.test.lib.hprof.model.*;
+import jdk.test.lib.hprof.parser.Reader;
+
+/*
+ * @test
+ * @summary Checks if -XX:+LimitPrimArrayContentInHeapDump works.
```
dbriemann: Needs update to the new naming : `LimitPrimitiveArrayContentInHeapDump`

dbriemann (commented):
One small change needed.

test/hotspot/jtreg/serviceability/HeapDump/PartialArrayContentTest.java
schmelter-sap: Done

dbriemann (dismissed):
lgtm

src/hotspot/share/services/heapDumper.cpp
```diff
@@ -2475,8 +2521,11 @@ void VM_HeapDumper::work(uint worker_id) {
 
   ResourceMark rm;
   // share global compressor, local DumpWriter is not responsible for its life cycle
-  DumpWriter segment_writer(DumpMerger::get_writer_path(writer()->get_file_path(), dumper_id),
-                            writer()->is_overwrite(), writer()->compressor());
+  // SapMachine 2026-05-06: Don't use segments if the dump is not parallel. This makes it
+  // possible to not use any disk space if dumping to a names pipe or a tty.
```
RealCLanger:
```suggestion
  // possible to not use any disk space if dumping to a named pipe or a tty.
```

RealCLanger (commented):
Found one minor spelling nit.

And a more general question: Don't you think this could be something to contribute to the OpenJDK? Like "space efficient generation of heap dumps" or such?

src/hotspot/share/services/heapDumper.cpp
schmelter-sap: Let's first get some practical experience with the feature to see if it would be a good idea to contribute it.

RealCLanger (approved):

RealCLanger (approved):

PR #2281, opened 2026-06-17
SapMachine (21) #2250: Add adjustments to heap dumps for the buildpack

### LINKED ISSUE #2250: Add adjustments to heap dumps for the buildpack

schmelter-sap:
This adds the same features for the heap dump changes in the buildpack done in 25+ to 21.

Since 21 doesn't have the problematic segemented heap dump, the fixes for that are omitted. The -XX:HeapDumpParallelism is kept for compatibility from 25+, but has no effect here (you cannot set the number of threads used for the heap dump in this implementation).

Additionally the checks for not triggering a "normal" segmented heap dump in the test was removed.

fixes #2250


src/hotspot/share/services/heapDumper.cpp
```diff
@@ -1531,6 +1553,24 @@ void DumperSupport::dump_prim_array(AbstractDumpWriter* writer, typeArrayOop arr
     return;
   }
 
+  // SapMachine 2026-06-16: If enabled, we don't dump the whole content of large arrays, but just the start
+  // and fill the rest with zeroes.
+  int fill_with_zero = 0;
+
+  if (LimitPrimitiveArrayContentInHeapDump) {
+      int limit = ArrayContentSizeLimitInHeapDump;
```
RealCLanger: the indentation looks wrong here

src/hotspot/share/services/heapDumper.cpp
```diff
@@ -543,6 +545,26 @@ void AbstractDumpWriter::write_id(u4 x) {
 #endif
 }
 
+// SapMachine 2026-06-16: Writes zeros to the buffer.
+void AbstractDumpWriter::write_zero(size_t len) {
+    assert(!_in_dump_segment || (_sub_record_left >= len), "sub-record too large");
```
RealCLanger: indentation

src/hotspot/share/services/heapDumper.cpp
```diff
@@ -1598,6 +1638,11 @@ void DumperSupport::dump_prim_array(AbstractDumpWriter* writer, typeArrayOop arr
     default : ShouldNotReachHere();
   }
 
+  // SapMachine 2026-06-16: Fill with zeros, if we don't dump the whole content of the array.
+  if (fill_with_zero > 0) {
+      writer->write_zero((u4)fill_with_zero * type_size);
```
RealCLanger: indentation

RealCLanger (changes requested):
Some indentation issues found

src/hotspot/share/services/heapDumper.cpp
schmelter-sap: Done

src/hotspot/share/services/heapDumper.cpp
schmelter-sap: Done

src/hotspot/share/services/heapDumper.cpp
schmelter-sap: Done

RealCLanger (approved):

PR #2308, opened 2026-06-30
SapMachine (27) #2250: Add adjustments to heap dumps for the buildpack

### LINKED ISSUE #2250: Add adjustments to heap dumps for the buildpack

RealCLanger:
Clean backport of 35ec0468c93e0dc27a09b799fd85b384017e2b74 from sapmachine to sapmachine27.

fixes #2250


schmelter-sap (approved):
Looks good.

Issue #1985, opened 2025-06-11
Fix Github Actions settings for new SapMachine branches


RealCLanger:
After branching off SapMachine release branches, e.g. 25, we need to configure Github Actions correctly.

PR #2325, opened 2026-07-03
SapMachine #1985: Fix sync-fork-checks concurrency group to be per-PR

RealCLanger:
The `sync-fork-checks.yml` workflow used `${{ github.ref }}` as the concurrency group key. For `pull_request_target` events, `github.ref` resolves to the base branch (e.g. `refs/heads/sapmachine`), not the PR branch — so all open PRs targeting the same base branch shared one concurrency slot, and a new run for any PR would cancel the currently-running one for any other PR.

Two fixes:
- Use `${{ github.event.pull_request.number }}` in the concurrency group key so each PR gets its own slot.
- Add a `cancelled()` cleanup step that marks any in-progress check runs for the commit as cancelled, preventing dangling `in_progress` check runs on the PR when the job is superseded.

fixes #1985

### DIFF
--- a/.github/workflows/sync-fork-checks.yml
+++ b/.github/workflows/sync-fork-checks.yml
@@ -38,7 +38,7 @@ on:
       - reopened
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


raeessap (approved):
LGTM

PR #2328, opened 2026-07-06
SapMachine (17) #2250: Add adjustments to heap dumps for the buildpack

### LINKED ISSUE #2250: Add adjustments to heap dumps for the buildpack

schmelter-sap:
Downport from sapmachine21.

fixes #2250


src/hotspot/share/services/heapDumper.cpp
```diff
@@ -555,6 +557,26 @@ void DumpWriter::write_id(u4 x) {
 #endif
 }
 
+// SapMachine 2026-06-16: Writes zeros to the buffer.
+void DumpWriter::write_zero(size_t len) {
+    assert(!_in_dump_segment || (_sub_record_left >= len), "sub-record too large");
```
RealCLanger: Here, again, the indentation is wrong and has 4 spaces. It should be 2.

RealCLanger (changes requested):

src/hotspot/share/services/heapDumper.cpp
schmelter-sap: Done.

schmelter-sap:
I've added explicit GC to use and young gen size to the test creating large arrays, since from time to time the test failed to create the big 2GB short array, even if the heap was 2.5 GB. Using a simple GC and limiting the young gen, we now make sure there is enough space in the old gen for the array.

RealCLanger:
> I've added explicit GC to use and young gen size to the test creating large arrays, since from time to time the test failed to create the big 2GB short array, even if the heap was 2.5 GB. Using a simple GC and limiting the young gen, we now make sure there is enough space in the old gen for the array.

Shouldn't that be ported to the other branches as well?

RealCLanger:
I fixed some more indentation.

RealCLanger (approved):

parttimenerd (approved):

MBaesken (approved):

PR #2357, opened 2026-07-23
SapMachine #2356: More logging in TestHiMemReportArgParsing

### LINKED ISSUE #2356: More logging in TestHiMemReportArgParsing
Since we see some problems here too, we should add more logging the test.

schmelter-sap:
Adds the same level of logging to the TestHiMemReportArgParsing test.

fixes #2356


RealCLanger (approved):
Looks good. PR test errors seem unrelated.

PR #2393, opened 2026-08-28
Sapmachine #2392: Add load average to vitals

### LINKED ISSUE #2392: Add load average to vitals
The load average is useful to see if a system is overloaded or not.

schmelter-sap:
This adds the load average to the vitals.

Since there are actually 3 load average values (taken over 1 minute, 5 minutes or 15 minutes) and the vitals table are already cryz big, this only adds one of the values which is most appropriate for the table (short term or long term). 

The load average is given in percent, with 100 percent meaning the system has exactly as many CPUs as runnable threads. This is done so we don't have to know the number of CPUs on the system to see if the system is overloaded.

In principle with the load average added. the "tr" column could be removed, since it is analog to the load average for a very small interval. Since this makes the value fluctuate wildly, not be suitable for the long term table and needs knowing the number of CPUs to interpret it, we could remove it. On the other hand, who knows who uses it and would not be happy to see it removed.
 
fixes #2392

### DIFF
--- a/src/hotspot/os/linux/vitals_linux_oswrapper.cpp
+++ b/src/hotspot/os/linux/vitals_linux_oswrapper.cpp
@@ -35,7 +35,6 @@
 #include <fcntl.h>
 #include <string.h>
 
-
 #define LOG_HERE_F(msg, ...)  { printf("[%d] ", (int)::getpid()); ::printf(msg, __VA_ARGS__); printf("\n"); fflush(stdout); }
 #define LOG_HERE(msg)         { printf("[%d] ", (int)::getpid()); ::printf("%s", msg); printf("\n"); fflush(stdout); }
 
@@ -371,6 +370,7 @@ const char* CGroups::_file_lim = nullptr;
 const char* CGroups::_file_limsw = nullptr;
 const char* CGroups::_file_slim = nullptr;
 const char* CGroups::_file_kusg = nullptr;
+double proc_scale_factor = 0.0;
 
 void OSWrapper::update_if_needed() {
 
@@ -394,6 +394,7 @@ ALL_VALUES_DO(RESETVAL)
 
     if (first_call) {
       log_trace(vitals)("Read /proc/meminfo: \n%s", bf.text());
+      proc_scale_factor = 100.0 / MAX2(1, os::processor_count());
     }
 
     // All values in /proc/meminfo are in KB
@@ -436,7 +437,7 @@ ALL_VALUES_DO(RESETVAL)
     _syst_cpu_st = values.steal;
     _syst_cpu_gu = values.guest + values.guest_nice;
 
-    // procs_running: this is actually number of threads running
+    // procs_running: this is running and runnable threads.
     // procs_blocked: number of threads blocked on real disk IO
     // See https://utcc.utoronto.ca/~cks/space/blog/linux/ProcessStatesAndProcStat
     // and https://lore.kernel.org/lkml/12601530441257@xenotime.net/#t
@@ -555,6 +556,22 @@ ALL_VALUES_DO(RESETVAL)
     }
   }
 #endif // __GLIBC__
+
+  if (bf.read("/proc/loadavg")) {
+    float l1, l5, l15;
+    if (sscanf(bf.text(), "%f %f %f", &l1, &l5, &l15) == 3) {
+      // Convert to relative percentage-based loads, where 100 percent
+      // means the number of runnable threads equals the number of CPUs.
+      value_t rl1 = (value_t) MAX2(0.0, MIN2(65535.0, proc_scale_factor * l1));
+      value_t rl5 = (value_t) MAX2(0.0, MIN2(65535.0, proc_scale_factor * l5));
+      value_t rl15 = (value_t) MAX2(0.0, MIN2(65535.0, proc_scale_factor * l15));
+      // We put the three values into one, since we want to display
+      // the longer averaged one in table with coarser resolution.
+      _syst_ldavg = (rl1 << 32) | (rl5 << 16) | rl15;
+    } else {
+      _syst_ldavg = INVALID_VALUE;
+    }
+  }
 
   first_call = false;
 
--- a/src/hotspot/os/linux/vitals_linux_oswrapper.cpp
+++ b/src/hotspot/os/linux/vitals_linux_oswrapper.cpp
@@ -570,6 +570,12 @@ ALL_VALUES_DO(RESETVAL)
       _syst_ldavg = (rl1 << 32) | (rl5 << 16) | rl15;
     } else {
       _syst_ldavg = INVALID_VALUE;
+      static bool traced = false;
+
+      if (!traced) {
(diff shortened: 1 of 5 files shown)


src/hotspot/os/linux/vitals_linux_oswrapper.cpp
```diff
@@ -556,6 +557,22 @@ ALL_VALUES_DO(RESETVAL)
...
+      // Convert to relative percentage-based loads, where 100 percent
+      // means the number of runnable threads equals the number of CPUs.
+      value_t rl1 = (value_t) MAX2(0.0, MIN2(65535.0, proc_scale_factor * l1));
+      value_t rl5 = (value_t) MAX2(0.0, MIN2(65535.0, proc_scale_factor * l5));
+      value_t rl15 = (value_t) MAX2(0.0, MIN2(65535.0, proc_scale_factor * l15));
+      // We put the three values into one, since we want to display
+      // the longer averaged one in table with coarser resolution.
+      _syst_ldavg = (rl1 << 32) | (rl5 << 16) | rl15;
+    } else {
+      _syst_ldavg = INVALID_VALUE;
```
MBaesken, line 572: Maybe add some debug UL logging for failures reading /proc/loadavg ?

src/hotspot/share/vitals/vitals.cpp
```diff
@@ -591,6 +595,32 @@ int DeltaMemorySizeColumn::do_print0(outputStream* st, value_t value,
   return 0;
 }
 
+int LoadAverageColumn::do_print0(outputStream* st, value_t value,
+    value_t last_value, int last_value_age, const print_info_t* pi) const {
+  if (value != INVALID_VALUE) {
+    value_t load_average;
+
+    // Use the minute resolution until an age of 2.5 minutes
+    // and 5 minute resulution until an age of 7.5 minutes.
```
MBaesken: resulution - little typo

MBaesken:
> Since there are actually 3 load average values (taken over 1 minute, 5 minutes or 15 minutes) and the vitals table are already cryz big, this only adds one of the values which is most appropriate for the table (short term or long term).

What does it look like?  
The column is 'Load average in percent' right? Would it be possible somehow to make it clear in the column name/header etc. that  it is a different resolution according to the comment ?

```
// Use the minute resolution until an age of 2.5 minutes
// and 5 minute resulution until an age of 7.5 minutes.
```

MBaesken (approved):
Some small comments/suggestions but seems to be okay.

MBaesken:
Or why not add all 3 values, if the table is large anyways ?

src/hotspot/os/linux/vitals_linux_oswrapper.cpp
```diff
@@ -394,6 +394,7 @@ ALL_VALUES_DO(RESETVAL)
 
     if (first_call) {
       log_trace(vitals)("Read /proc/meminfo: \n%s", bf.text());
+      proc_scale_factor = 100.0 / MAX2(1, os::processor_count());
```
MBaesken, line 397: Lazy init on first call; that's not wrong in most/many cases but do we want to support changing number of CPUs like hotplug mechanisms? (but not sure how well this is supported across the whole codebase, so probably we are not worse than other coding)

MBaesken (dismissed):

schmelter-sap:
It is currently not possible to change the column header for the different tables. And it would not be easy to change that.

Currently the type of load average used is determined by the resolution of the table. In the short time table the samples are 10 seconds apart, so seeing the 15 minute load average is not really that useful compared to the 1 minute average. On the other hand, in the long term table the sample interval is 1 hour and there the 15 minute average better represents the 1 hour interval than the shorter average.

I think the main question to answer with the load average is: Was the system in the sample interval overloaded? And for this one should look at the load average type which is nearest to the sample interval.

If you're really interested in the type you see, one could annotate the type in the non-raw output. E.g. for a 120 percent load instead of using "120" one could use "120/1", 120/5" or "120/15" .

ansteiner:
I think tr and tb was used as these are well known metrics from the system e.g. from vmstat. I used them in the past sometimes. Not sure if others rely on these.

src/hotspot/os/linux/vitals_linux_oswrapper.cpp
```diff
@@ -556,6 +557,22 @@ ALL_VALUES_DO(RESETVAL)
...
+    float l1, l5, l15;
+    if (sscanf(bf.text(), "%f %f %f", &l1, &l5, &l15) == 3) {
+      // Convert to relative percentage-based loads, where 100 percent
+      // means the number of runnable threads equals the number of CPUs.
+      value_t rl1 = (value_t) MAX2(0.0, MIN2(65535.0, proc_scale_factor * l1));
+      value_t rl5 = (value_t) MAX2(0.0, MIN2(65535.0, proc_scale_factor * l5));
+      value_t rl15 = (value_t) MAX2(0.0, MIN2(65535.0, proc_scale_factor * l15));
+      // We put the three values into one, since we want to display
+      // the longer averaged one in table with coarser resolution.
+      _syst_ldavg = (rl1 << 32) | (rl5 << 16) | rl15;
```
ansteiner, line 570: How meaningful will this be in the "Last 60 minutes section" with 10 seconds interval?

schmelter-sap: In the 10 second interval table the 1 minute average will be used, which covers 6 times the sample interval. While not ideal, it is easy to see the general change of the load compared to the previous sample.

schmelter-sap:
> I think tr and tb was used as these are well known metrics from the system e.g. from vmstat. I used them in the past sometimes. Not sure if others rely on these.

OK, I will not touch it.

src/hotspot/os/linux/vitals_linux_oswrapper.cpp
schmelter-sap: One could use os::active_processor_count(), but this calls sysconf(_SC_NPROCESSORS_ONLN) which uses get_nprocs_conf() to get the value. And the man page for get_nprocs_conf() warns that this is an expensive fdunction to call.

src/hotspot/os/linux/vitals_linux_oswrapper.cpp
schmelter-sap: Done

src/hotspot/share/vitals/vitals.cpp
schmelter-sap: Fixed.

src/hotspot/os/linux/vitals_linux_oswrapper.cpp
MBaesken: Okay, if you think it is expensive then keep as it is.

src/hotspot/os/linux/vitals_linux_oswrapper.cpp
MBaesken: Thanks!

PR #2409, opened 2026-09-11
SapMachine #2392: Add load average to vitals

### LINKED ISSUE #2392: Add load average to vitals

schmelter-sap:
The change adds a load average value for all operating systems.

It only uses one value as the load average which tries to be specific for the sample interval of vitals. For the long term values the load average is defined as the short term values averaged over the sample interval of the long term table.

This change adds the ability to use different values for entries in the long term table. Until now that value in the long term table is just the same as in the short term table for the given timestamp. This is OK for some types of values but, for others a value which more represents the sample interval of the long term table might be better.

We also split up the initialization of the vitals system into two phases, since we don't want to start the sampler thread until the himem report system is initialized too (otherwise there is a race condition).

And finally some traces were improved.

fixes #2392

### DIFF
--- a/src/hotspot/os/windows/vitals_windows.cpp
+++ b/src/hotspot/os/windows/vitals_windows.cpp
@@ -101,7 +101,7 @@ bool platform_columns_initialize() {
       define_column<MemorySizeColumn>("system", nullptr, "avail-phys", "Amount of physical memory currently available.", true, MIN);
 
   g_col_system_load_average =
-    define_column<MemorySizeColumn>("system", nullptr, "la", "Load average of system in percent.", has_loadavg, MAX);
+    define_column<PlainValueColumn>("system", nullptr, "la", "Load average of system in percent.", has_loadavg, MAX);
 
   // PROCESS_MEMORY_COUNTERS_EX WorkingSetSize
   g_col_process_working_set_size =
--- a/src/hotspot/os/windows/vitals_windows.cpp
+++ b/src/hotspot/os/windows/vitals_windows.cpp
@@ -121,24 +121,15 @@ static void set_value_in_sample(Column* col, Sample* sample, value_t val) {
   }
 }
 
-static double get_load_average() {
+static value_t get_load_average() {
   if (!has_loadavg) {
-    return -1.0;
+    return INVALID_VALUE;
   }
 
   return get_load_average_impl(false);
 }
 
 void sample_platform_values(Sample* sample, Sample* long_term_sample) {
-  double load_avg = get_load_average();
-
-  add_load_average(load_avg);
-  set_value_in_sample(g_col_system_load_average, sample, load_avg);
-
-  if (long_term_sample != nullptr) {
-    set_value_in_sample(g_col_system_load_average, sample, get_long_term_load_average());
-  }
-
   MEMORYSTATUSEX mse;
   mse.dwLength = sizeof(mse);
   if (::GlobalMemoryStatusEx(&mse)) {
@@ -152,6 +143,8 @@ void sample_platform_values(Sample* sample, Sample* long_term_sample) {
     set_value_in_sample(g_col_process_working_set_size, sample, cnt.WorkingSetSize);
     set_value_in_sample(g_col_process_commit_charge, sample, cnt.PagefileUsage);
   }
+
+  set_load_average(g_col_system_load_average, get_load_average(), sample, long_term_sample);
 }
 
 } // namespace sapmachine_vitals
--- a/src/hotspot/os/windows/vitals_windows.cpp
+++ b/src/hotspot/os/windows/vitals_windows.cpp
@@ -23,10 +23,12 @@
  *
  */
 
+#include "logging/log.hpp"
 #include "runtime/os.hpp"
 #include "utilities/debug.hpp"
 #include "utilities/globalDefinitions.hpp"
 #include "vitals/vitals_internals.hpp"
+#include "pdh_interface.hpp"
 
 #include <psapi.h>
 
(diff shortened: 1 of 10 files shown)


MBaesken:
Ohh nooo, AIX fails to compile! But I see a commit dealing with AIX so I wonder if this is already fixed??

ansteiner:
> Ohh nooo, AIX fails to compile! But I see a commit dealing with AIX so I wonder if this is already fixed??

I think there is no getloadavg() on AIX. Isn't it? Have to use perfstat_cpu_total.

schmelter-sap:
> > Ohh nooo, AIX fails to compile! But I see a commit dealing with AIX so I wonder if this is already fixed??
> 
> I think there is no getloadavg() on AIX. Isn't it? Have to use perfstat_cpu_total.

There is an implementation for os::loadavg() on aix at least. And that's what I wanted to use, but the "os" part is missing.

MBaesken:
The column is named
'Load average of system in percent'
but on some platforms
'Load average in the sample interval in percent'
is this intentional?
Is it NOT of the system in the second case ?

MBaesken:
From the info I find, the PDH stuff (e.g. PROCESSOR_TIME) might be localized.
Seems this is the reason the OpenJDK codebase uses indices and not those 'paths', there are some potentially useful comments and code here

https://github.com/openjdk/jdk/blob/af4116009d8567aab8f23a04fc79e801b52f60d4/src/hotspot/os/windows/os_perf_windows.cpp#L71
https://github.com/openjdk/jdk/blob/af4116009d8567aab8f23a04fc79e801b52f60d4/src/jdk.management/windows/native/libmanagement_ext/OperatingSystemImpl.c#L292

(there are also english-name counter functions, maybe those help too to get around issues on non-english Windows https://learn.microsoft.com/en-us/windows/win32/api/pdh/nf-pdh-pdhaddenglishcounterw)

schmelter-sap:
> From the info I find, the PDH stuff (e.g. PROCESSOR_TIME) might be localized.

According to the SAPJVM code this doesn't matter from Vista onward. But I can test it on my machine.

MBaesken:
Maybe the comments in os_perf_windows about this localization stuff
https://github.com/openjdk/jdk/blob/af4116009d8567aab8f23a04fc79e801b52f60d4/src/hotspot/os/windows/os_perf_windows.cpp#L42
are a little outdated, who knows.

Fun side fact , if the docu is correct, the language neutral helper  PdhAddEnglishCounterW   https://learn.microsoft.com/en-us/windows/win32/api/pdh/nf-pdh-pdhaddenglishcounterw
_requires_ Vista or higher. So we would have a helper on OS , where the helper is not needed any more !??

schmelter-sap:
> The column is named
> 'Load average of system in percent' ...

Ups, that is the initial description. I later changed to do indicate it covers the sample interval of the table.

schmelter-sap:
> So we would have a helper on OS , where the helper is not needed any more

Good question. It doesn't seem to make sense. I will switch to PdhAddEnglishCounterW.

Edit:
Or not. I'll look into the counter-base access, since otherwise I would have to touch the pdh_interface code and I want to keep this change to SapMachine code only.

schmelter-sap:
retest this please

schmelter-sap:
retest this please

src/hotspot/share/vitals/vitals.cpp
```diff
@@ -962,18 +969,24 @@ static SampleTables* g_all_tables = nullptr;
...
   DEBUG_ONLY(sample->set_num(-1);)
   sample_jvm_values(sample, avoid_locking);
-  sample_platform_values(sample);
+  sample_platform_values(sample, long_term_sample);
 }
 
 class SamplerThread: public NamedThread {
 
   Sample* _sample;
+  Sample* _long_term_sample;
```
ansteiner, line 989: _long_term_sample is not initialized in the SamplerThread member initializer list like _sample

ansteiner (dismissed):
not really an issue ;-)

src/hotspot/share/vitals/vitals.cpp
schmelter-sap: Moved the allocation to the initialization list.

ansteiner (dismissed):

RealCLanger (dismissed):

RealCLanger:
I think you could update the copyright years, though.

MBaesken:
queue_lengt_counter_name variable  - seems to be a typo, missing 'h'.

src/hotspot/os/windows/vitals_windows.cpp
```diff
@@ -23,27 +23,140 @@
...
+
   g_col_system_memoryload =
       define_column<PlainValueColumn>("system", nullptr, "mload", "Approximate percentage of physical memory that is in use.", true, MAX);
 
   // MEMORYSTATUSEX ullAvailPhys
   g_col_system_avail_phys =
       define_column<MemorySizeColumn>("system", nullptr, "avail-phys", "Amount of physical memory currently available.", true, MIN);
+
+  g_col_system_load_average =
+    define_column<PlainValueColumn>("system", nullptr, "la", "Load average in the sample interval in percent.", has_loadavg, MAX);
```
MBaesken: We have here 'Load average in the sample interval in percent.' wth a '.' at the end.  The other define_column calls seem to miss the point.  No big deal but probably not desired.

Update: seems we have this in _vitals_windows.cpp‎_  for some other columns too; not sure why but the 'style' was taken from existing code so probably okay.

MBaesken:
Maybe add the info to the related issue
https://github.com/SAP/SapMachine/issues/2392
that the values of **VitalsLongTermSampleIntervalMinutes** was changed from 60 to 10, and **VitalsLongTermTableDays** from 14 to 7, to make  it a bit more transparent.

MBaesken (approved):

src/hotspot/os/windows/vitals_windows.cpp
schmelter-sap: Remove the dot for windows.

PR #2421, opened 2026-09-21
SapMachine (26) #2392: Add load average to vitals

schmelter-sap:
Downport from head.

fixes #2392


RealCLanger:
Sorry to say, but 26 is not needed any more 😉

